PR #10, opened 2017-12-14
Windows host app Makefile

arcnmx:
Some changes that enable building of the host app using mingw-w64. I don't have VS to test building in Windows but it still should work?

Haven't tried compiling with nvfbc, and added a define to leave it out. Maybe we could add the headers in the `vendor` folder?

Note: the newer DXGI interfaces don't seem to be included in the latest released version of mingw-w64. On Arch you can install `mingw-w64-headers-git` and `mingw-w64-crt-git` from AUR to get them.

### DIFF
--- a/host/Capture/DXGI.cpp
+++ b/host/Capture/DXGI.cpp
@@ -51,7 +51,7 @@ bool DXGI::Initialize(CaptureOptions * options)
   status = CreateDXGIFactory1(__uuidof(IDXGIFactory1), (void **)(&m_dxgiFactory));
   if (FAILED(status))
   {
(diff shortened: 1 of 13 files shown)


gnif:
Excellent work! Thanks. I will verify this is all good in VS and if so I will merge it.

gnif:
Thanks! I had to adjust a few things to make it build in VC++, one specific of note was `__uidof` in 06136afd, can you please verify this still works under mingw?

arcnmx:
It still looks good! Builds without modification from master and a quick test has the binary working as well as it did before.

gnif:
Great, thanks again!

PR #14, opened 2017-12-15
Add borderless fullscreen option

jack-karamanian:
This PR adds the option for a borderless fullscreen window.
This allows the window to display over UI elements such as the Gnome 3 top bar.
In the case of Gnome 3, the window will display at full resolution with this option, instead of shrinking the window to fit within the space below the top bar.

Currently this option can be specified by `-F` on the command line.

### DIFF
--- a/client/main.c
+++ b/client/main.c
@@ -67,6 +67,7 @@ struct AppParams
   bool         allowResize;
   bool         keepAspect;
   bool         borderless;
+  bool         fullscreen;
   bool         center;
   int          x, y;
   unsigned int w, h;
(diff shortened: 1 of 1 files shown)


gnif:
Brilliant, thanks!

artizirk:
Seems to work under Arch Linux with Gnome 3.26 when switching to another desktop the client will be minimized and when swithing back to that desktop i have to do a quick alt-tab to bring the client back to focus and fullscreen

artizirk:
This behaviour seems to be intentional by libsdl and it looks like it is not configurable.

jack-karamanian:
Indeed that does work.

PR #19, opened 2017-12-15
[host] simplify dll loading

jmossman:
The code that's there now is re-implementing what LoadLibrary does internally under Wow64. The ifdef for `NVFBC_LIBRARY_NAME` is the only thing that's actually needed.


gnif:
Thanks, this was written based on a NVidia sample, really didn't think twice about it.

gnif:
Now I see why you wanted this! Google cache shows you are the idiot that posted the nvidia-enabler code that just turned up. Reverting this to avoid abuse like yours.

PR #24, opened 2017-12-16
NvFBC on mingw

arcnmx:
Fixes to get the NvFBC capture driver compiling on Linux. `make -C host CONFIG_CAPTURE_NVFBC=1`

Note: Requires extracting the `NvFBC` headers from the `inc` directory of the capture SDK into the `vendor` folder. The particular header also has `Windows.h` incorrectly cased and needs adjustment there. It would be nice if we could just include it in the repo, or rewrite it if there are copyright concerns about using it verbatim, but I'll leave that out for now.


gnif:
Looks good thanks, and yes we can not include the NVIDIA headers in the package, it would breach the license agreement.

PR #28, opened 2017-12-21
Mingw compile fixes

arcnmx:
Not sure if `restart` not being used was intentional or not, I haven't had a chance to try the build yet.


gnif:
Confirmed, `restart` was deprecated. Thanks again.

PR #36, opened 2018-01-13
Allow edge-of-screen mouse position sync/fix/jump to work in nvfbc mode

zetaPRIME:
Not exactly sure what that feature is called, but naturally it only actually works if the client knows where the client's mouse position is; this patch adds that information to nvfbc's output to bring it into parity with DXGI.

### DIFF
--- a/host/Capture/NvFBC.cpp
+++ b/host/Capture/NvFBC.cpp
@@ -28,6 +28,8 @@ using namespace Capture;
 #include "common/memcpySSE.h"
 #include "Util.h"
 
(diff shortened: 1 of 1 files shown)


gnif:
I am sorry but this is not the correct way to deal with this. NvFBC supports mouse cursor and positional streaming instead of compositing the cursor onto the image. LG needs to be updated to make use of this just as DXGI does.

When I return from break I and DXGI stabilizes I will return my focus to NvFBC and implement this.

PR #38, opened 2018-01-14
Ignore SDL_MOUSEBUTTONDOWN events incompatible with SPICE

mcd1992:
The SPICE protocol only supports a standard 3 button PS/2 virtual mouse. Right now SDL is passing through mouse 4/5 + which SPICE is interpreting as scroll up / down. Drop these events to avoid weird scrolling on mouse 4/5.

### DIFF
--- a/client/main.c
+++ b/client/main.c
@@ -587,7 +587,8 @@ int eventThread(void * arg)
       }
 
       case SDL_MOUSEBUTTONDOWN:
-        if (event.button.button > 3) // The SPICE protocol doesn't support more than a standard PS/2 3 button mouse
+        // The SPICE protocol doesn't support more than a standard PS/2 3 button mouse
+        if (event.button.button > 3)
           break;
         if (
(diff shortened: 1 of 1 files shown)


mcd1992:
Updated, thanks.

JaneSmith:
Wait, so this means that the built-in input handling will never be able to support more than three mouse buttons? I really like the Spice input client, but for gaming that sounds really, really bad. Are there any solutions to this?

mcd1992:
The best solution right now would be to passthrough a keyboard/mouse to the VM and use them. As for solutions that allow passthrough of the host devices I'm not aware of any. Might be worth digging into how steam in-home streaming uses a SDL based streaming client and manages to passthrough all keyboard/mouse events with low latency.

Another interesting concept might be to create another shared memory region (or re-use the existing one) and put mouse/keyboard events in it then have the looking-glass-host.exe parse them and pass them as a virtual mouse/keyboard. Not sure how that would work on the windows end though.

PR #40, opened 2018-01-28
[client] Makefile: fix multi-job race condition

Xiretza:
make fails when using `-j` because xlib-shim.so can be targeted before the binary, in which case bin/ doesn't exist yet.


gnif:
Duplicate of #43

PR #43, opened 2018-01-28
[client] fix build error when using parallel make

arcnmx:
`$(BIN)` may not exist at the time that `xlib-shim.so` is built.


gnif:
Thanks.

Rather then adding another `mkdir` please create another task that creates the directory that both the executable and library depend on.

PR #44, opened 2018-01-28
[host] mingw-w64 compile fixes

arcnmx:
Not sure if you'd want to merge this as-is, I can split it up into a few commits if it makes sense to do so. The main issue here is that mingw headers are missing some of the newer media foundation APIs.

Also `min` was moved out of the util header because `#define min` conflicts with some libc/libstdc++ headers if they're included after it. Might be nice to just define a `LG_MIN` if it's needed in multiple parts of the code?

### DIFF
--- a/host/Capture/DXGI.cpp
+++ b/host/Capture/DXGI.cpp
@@ -386,9 +386,9 @@ bool DXGI::InitH264Capture()
     return false;
   }
 
-  m_mfTransform->ProcessMessage(MFT_MESSAGE_COMMAND_FLUSH         , NULL);
-  m_mfTransform->ProcessMessage(MFT_MESSAGE_NOTIFY_BEGIN_STREAMING, NULL);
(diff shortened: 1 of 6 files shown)


gnif:
Thanks, please split up the functional changes into separate commits.

Rather then move `min` can you use an ifdef for `min` rather than `mingw`? Ie: `if !defined(min)`

arcnmx:
While I can change the condition to that, it still needs to be either moved or renamed.

gnif:
Ok, fair enough. Please just split this up into separate commits as it addresses several things at once.

  * printf changes
  * DXGI changes
  * Util (min) change

arcnmx:
Changes made, with `min` kept in Util.h but renamed to `LG_MIN`. Also restored its use in DXGI.cpp.

gnif:
Excellent, thanks!

PR #45, opened 2018-01-29
[client] fix event loop sleep

arcnmx:
Moves the sleep so that it (only) gets hit when there are no events in the queue. Previously it would only throttle events after they occur, which is rare and had no impact on the 100%+ cpu usage.


gnif:
Doh!, Good catch :+1:

PR #46, opened 2018-02-01
[client]: ignore Alt-F4 in ignoreQuit mode

alama:
Give a Hint to SDL2 that we are ignoring ALT-F4

### DIFF
--- a/client/main.c
+++ b/client/main.c
@@ -786,6 +786,11 @@ int run()
     )
   );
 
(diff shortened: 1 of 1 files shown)


alama:
so, should I change `params.fullscreen` as well?

gnif:
Ah yes, perfect, thanks! That one slipped by.

PR #48, opened 2018-03-05
Stepped mouse movement for virtio and related tweaks

zetaPRIME:
As you may have noticed, having a virtio mouse present breaks "edge jumping" on Spice; if the cursor needs to move more than a certain distance, no movement occurs. This is because the virtio mouse driver uses a signed byte for each axis, resulting in a silent failure if either axis on a single movement packet is out of that range.

The first commit in this series adds a function that applies the necessary mouse movement in steps of up to +-120px on each axis, allowing functionally uncapped mouse movement. I didn't bother to add an early-out by magnitude, as it wouldn't save any significant amount of calculation over simply going through a single loop (and it shouldn't be enough extra work to be any sort of issue anyway).

The second commit expands the edge jump/cursor realignment code to allow it to function even without a reported position from the guest, by first moving up and left by the size of the screen (catching it in the upper left corner), then moving back down by its target position. This of course assumes a regular workspace shape, but since only a single screen is supported this shouldn't be an issue.

The third commit simply upsizes the mouse queue to accommodate the potential number of steps required to traverse to the upper left corner and back on higher resolutions.

### DIFF
--- a/client/main.c
+++ b/client/main.c
@@ -420,6 +420,27 @@ static inline const uint32_t mapScancode(SDL_Scancode scancode)
   return ps2;
 }
 
+// virtio mouse driver only works with move events within the range of a signed byte; in addition,
+// other mouse drivers are not used by spice when virtio mouse is present
+// however, virtio is currently far more stable than emulated PS/2 or USB
+#define VIRTIO_MAX_STEP 120
+bool smart_mouse_motion(int32_t x, int32_t y)
+{
+  int px = 0; int py = 0; // "parity" (sign) x/y (s was taken)
+  if (x != 0) { if (x < 0) px = -1; else px = 1; }
+  if (y != 0) { if (y < 0) py = -1; else py = 1; }
+  
+  while (x != 0 || y != 0) { // move mouse by maximum step at a time
+    int sx = x; int sy = y; // step x/y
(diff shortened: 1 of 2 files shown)


gnif:
Looks good but I believe your reasoning on the signed/unsigned byte is flawed, you are working around a bug in the spice server which should know to implement stepped movement if the driver doesn't support said movement values. This problem should be addressed in the spice server in Qemu, not Looking Glass.

zetaPRIME:
Ideally yes, but judging by the lack of progress on the PS/2 emulation bugs, that side of things is rather slow-moving on these (rather specific, I might add) issues. Looking Glass, on the other hand, is small and specific enough to be *far* more "agile" about improving the experience around its specific use case. If nothing else, this can be a stopgap measure on a command-line flag, and removed once qemu upstream fixes the underlying issue and makes it irrelevant.

gnif:
QEMU already implements stepped movements when the delta is too large, but due to a race condition in QEMU the stream would get corrupted. The following patches to QEMU fix this:

* https://lists.gnu.org/archive/html/qemu-devel/2018-05/msg01431.html
* https://lists.gnu.org/archive/html/qemu-devel/2018-05/msg01430.html

PR #51, opened 2018-03-19
Added a no video and a no framethread option (as argument)

Alkeryn:
Because both drastically affected performance in some application
especially input lag

for example in osu mania i went from 1ms of latency to 0.35ms which can affect scores on a rythm game

also put it in the help so you can read it
and the no framethread option kills also video for some reason, still have to look up why
also if you ask why not using virt-viewer
because it have terrible keyboard latency and mouse for me at least
and looking glass is the most working spice client i found yet
expept that it don't lock keyboard well (yet)

### DIFF
--- a/client/main.c
+++ b/client/main.c
@@ -95,13 +95,15 @@ struct AppParams
   bool         scaleMouseInput;
   bool         hideMouse;
   bool         ignoreQuit;
+  bool         noVideo;
+  bool         FrameThread;
 
   bool         forceRenderer;
   unsigned int forceRendererIndex;
   RendererOpts rendererOpts[LG_RENDERER_COUNT];
 };
 
-struct AppState  state;
+struct AppState  state; 
 struct AppParams params =
 {
   .configFile       = "/etc/looking-glass.conf",
@@ -123,7 +125,9 @@ struct AppParams params =
   .scaleMouseInput  = true,
   .hideMouse        = true,
   .ignoreQuit       = false,
-  .forceRenderer    = false
+  .forceRenderer    = false,
+  .noVideo          = false,
+  .FrameThread      = true,
 };
 
 static inline void updatePositionInfo()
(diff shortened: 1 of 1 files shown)


gnif:
There are several problems I have with this change set.

- Disabling the frame thread will cause the host to hang waiting for the client to ack the frame, see https://github.com/gnif/LookingGlass/blob/fd8d4d3d383f59742c7db782c140e5a158fbd0a0/client/main.c#L314
- A better/simpler approach would be to implement a NULL renderer, you could then also provide visual feedback instead to the user that it's headless. see: https://github.com/gnif/LookingGlass/tree/master/client/renderers
- This will not reduce latency as comments imply, it could actually cause latency due to hanging up the host application in a spin lock while waiting on the client to ack the frame.

If you want a plain spice input client you would be better off using the spice library and implementing it as an entirely separate application. https://github.com/gnif/LookingGlass/tree/master/client/spice

Alkeryn:
Yeah i saw the framethread thing after pushing
i just dived into your code i didn't understood everything yet ^^
though from bench i got lower latency by disabling this way for some reasons
I just saw it at first and disabled it by curiosity,and it resulted in better bench that's it
though i should try the way you talked since it is more optimised, thanks ^^

will have to read a bit more since i don't know how spice works really
and i didn't red more than main.c yet, but thanks for the information :)

Alkeryn:
Though is it okay to diasable the render thread ? or it also result in a spin lock ?

Alkeryn:
Though now that i think of
a spice client only option would be the better thing
rahter than that hacky commit
will read more and do it later

gnif:
I suggest you just take a copy of the spice directory and implement your own client, I wrote it to be modular for exactly this reason.

Alkeryn:
Thank you, really mate your spice implementation is top notch :)

Alkeryn:
Okay so i found something better than spice
evdev passthrough
so it have perfect locking
0 latency and all
so my fork is kinda useless now
exept for the seamless part
so there i will remove the no framthread option
and work on a null renderer
but yeah the only reason now is for seamless mouse
and even i disbaled the server from vm so yeh

gnif:
I am sorry but this does not at all sit within what Looking Glasse's goals are, it will not be accepted in any form.

Alkeryn:
Exactly ^^

Issue #82, opened 2018-06-25
windows-host-v11 stops running immediatly


caHocke:
### Required information

Host CPU: Intel Core i7-4770 CPU @ 3.9GHz
Host GPU: Mesa DRI Intel(R) Haswell Desktop
Guest GPU: NVIDIA GeForce 780ti
Host Kernel version: x86_64 Linux 4.9.0-6-amd64
Host QEMU version: version 2.8.1(Debian 1:2.8+dfsg-6+deb9u3)

I tried upgrading from looking-glassv10 to v11. 
The windows host immediately stops running with the only output being:

> [I]     CaptureFactory.h:83   | CaptureFactory::DetectDevice   | Trying DXGI
> [I]     CaptureFactory.h:86   | CaptureFactory::DetectDevice   | Using DXGI
> [I]          Service.cpp:132  | Service::InitPointers          | Total Available :  16 MB
> [I]          Service.cpp:133  | Service::InitPointers          | Max Cursor Size :   1 MB
> [I]          Service.cpp:134  | Service::InitPointers          | Max Frame Size  :   7 MB
> [I]          Service.cpp:135  | Service::InitPointers          | Cursor          : 000001EB0D320080 (0x00000080)
> [I]          Service.cpp:136  | Service::InitPointers          | Frame 1         : 000001EB0D420080 (0x00100080)
> [I]          Service.cpp:137  | Service::InitPointers          | Frame 2         : 000001EB0DBA0000 (0x00880000)

gnif:
How do you mean, "stops running", does the process terminate? and if so, is it immediate or when you run the client?
Where is your client log?
Did you upgrade the client?
Are you using a master build or a tagged release?
What resolution is your guest running at?

caHocke:
I did not get to start the client, I just started the host application and after generating htis output it stopped running. I could not find it in the Taskmanger. I can't find any log files. I assume they get generated at the same spot the .exe is located? I started it at various  locations to make sure it has nothing to do with read/write protection.

Lessaj:
What are you emulating as the guest CPU? Is it passthrough or some other kind such as core2duo?

gnif:
The application requires SSE2 instructions, at current it will silently fail if you do not have them available to the guest. Please ensure your guest is running with a CPU that supports them, either use core2duo or better, such as host-passthrough.

caHocke:
I changed the cpu mode to host-passthrough and it works now. Thank you :+1:

Issue #76, opened 2018-06-03
Alpha 11 breaks compatibility with my Windows 10 guest


Djhg2000:
Host CPU: AMD Ryzen 2400G
Host GPU: AMD Ryzen 2400G
Guest GPU: Sapphire Radeon RX 580
Host Kernel version: 4.17-rc7
Host QEMU version: version 2.12.0 (Debian 1:2.12+dfsg-3)

LG host behaves like the SSE issue is described, but SSE seems to be enabled (PCSX2 works with SSE2 enhancements enabled). Is there any more reliable way to test if SSE is the issue rather than just "LG silently crashes"? Guest is running on core2duo vcpus for compatibility reasons, I might be able to set up a test machine with host-passthrough vcpus later today.

I managed to capture a screenshot just before the crash [here](https://imgur.com/a/VwNavqL).

Alpha 10 works flawlessly and performs well at 1080p, if this turns out to be SSE related is there any chance we can get back the generic memcpy() implementation behind a compile flag? I don't have a Windows compatible build environment and I doubt I'll have time to set it up until next weekend.

No `looking-glass-host.dmp` file is created which I presume means a hard crash.

Lessaj:
core2duo is not compatible, it does not have the extensions needed. You will need to manually add them if you want to continue to use that model but that may break your compatibility with other applications. In my case, I was using core2duo because otherwise Asus GPU Tweak would cause a BSOD, I was using that to control the external fan headers on my card but I could not come up with an extension list that allowed both the LG host and Asus GPU Tweak to work without a BSOD, I ended up using host-passthrough and bought a y-splitter to control both fans from one header on the mobo.

You can add features under the <cpu> section like this. I kind of threw away the notes I made when I was trying this all out but I think these are the policies you need to add, you can see all of them in `/usr/share/libvirt/cpu_map.xml`.

      <feature policy='require' name='avx'/>
      <feature policy='require' name='x2apic'/>
      <feature policy='require' name='xgetbv1'/>
      <feature policy='require' name='xsave'/>
      <feature policy='require' name='xsavec'/>
      <feature policy='require' name='xsaveopt'/>

The error BTW is "illegal instruction", I ran the host code in debug mode in Visual Studio to find the instruction set(s) that allowed it to continue running without that error.

namidairo:
> Alpha 10 works flawlessly and performs well at 1080p, if this turns out to be SSE related is there any chance we can get back the generic memcpy() implementation behind a compile flag?

[...]

PR #85, opened 2018-07-05
Check AVX Support

eduxstad:
This commit adds a function which checks for AVX support in the OS/CPU before the assembly is run (which requires it). The function was copied from the answer [here](https://stackoverflow.com/a/22521619) which is expanded [here](https://insufficientlycomplicated.wordpress.com/2011/11/07/detecting-intel-advanced-vector-extensions-avx-in-visual-studio/). Currently the code only outputs a warning instead of ending the program, because I don't want to risk closing the program if the check is not working. This was referenced in #82 and #76.

### DIFF
--- a/host/main.cpp
+++ b/host/main.cpp
@@ -40,6 +40,7 @@ void doLicense();
 
 bool consoleActive = false;
 void setupConsole();
+bool avxSupport();
 
 struct StartupArgs
 {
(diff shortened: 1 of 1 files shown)


eduxstad:
I'm afraid I don't understand. Isn't vmovaps (used in memcpySSE) an AVX instruction? Probably doing something wrong?

gnif:
Done, see a72ad4e46cbd47a56514b256ab01a97b54e81add
Minimum requirements are now SSE2

eduxstad:
Alright, gonna close this pr because it's now basically irrelevant. I'll open a new one if I get around to rewriting the check for sse2.

PR #89, opened 2018-07-31
[client] Fix segfault when envar XDG_SESSION_TYPE isn't set

noneucat:
Minimal systems in cases may not have XDG_SESSION_TYPE set at all, causing the program to segfault at the `strcmp`. This commit sets XDG_SESSION_TYPE to `unspecified` (according to https://www.freedesktop.org/software/systemd/man/pam_systemd.html) if it is not defined in the environment.


YvanDaSilva:
LGTM.
Well spotted.

Issue #107, opened 2018-12-16
OpenGL renderer errors with `glGenBuffers = 1281` on a12


ananace:
### Required information

Host CPU: i7-7700K
Host GPU: Radeon RX 580
Guest GPU: Nvidia GeForce GTX 1070
Host Kernel version: 4.18.19
Host QEMU version: 3.0.0

This is with Xorg version: 1.20.3

Disabling any/all of the renderer options has no effect, still giving the same error.

```
$ ./looking-glass-client  -g opengl
[I]               main.c:757  | run                            | Looking Glass (a12-0-g507732587e)
[I]               main.c:758  | run                            | Locking Method: Atomic
[I]               main.c:806  | run                            | Trying forced renderer
[I]               main.c:751  | try_renderer                   | Using Renderer: OpenGL
[I]              spice.c:159  | spice_connect                  | Remote: 127.0.0.1:5900
[I]               main.c:961  | run                            | Waiting for host to signal it's ready...
[I]              spice.c:367  | spice_on_common_read           | notify message: keyboard channel is insecure
[I]             opengl.c:469  | opengl_render_startup          | Vendor  : X.Org
[I]             opengl.c:470  | opengl_render_startup          | Renderer: Radeon RX 580 Series (POLARIS10, DRM 3.26.0, 4.18.0-sabayon, LLVM 7.0.0)
[I]             opengl.c:471  | opengl_render_startup          | Version : 4.4 (Compatibility Profile) Mesa 18.2.5
[I]             opengl.c:483  | opengl_render_startup          | Using GL_AMD_pinned_memory
[I]               main.c:970  | run                            | Host ready, starting session
[I]             opengl.c:896  | configure                      | Using decoder: NULL
[E]             opengl.c:863  | _check_gl_error                | 955: glGenBuffers = 1281 (invalid value)
[I]             opengl.c:896  | configure                      | Using decoder: NULL
```

Issue #101, opened 2018-10-27
935: glGenBuffers = 1281 (invalid value)


Topis:
### Required information

Host CPU: Ryzen 2700X
Host GPU: Radeon R7 240
Guest GPU: GeForce GTX 1060
Host Kernel version: 4.18.0-16.1-liquorix-amd64
Host QEMU version: 3.0.0

Commit https://github.com/gnif/LookingGlass/commit/961312716222e703704988980e3de190e3a613b2 breaks the client on Wayland for me. I get the following error and a black screen.

[I]               main.c:735  | run                            | Looking Glass (a11-98-g2692ccc7b3-dirty
)                                                                                                       
[I]               main.c:736  | run                            | Locking Method: Atomic                 
[I]               main.c:751  | run                            | Wayland detected                       
[I]               main.c:757  | run                            | SDL_VIDEODRIVER has been set to wayland
[I]               main.c:729  | try_renderer                   | Using Renderer: OpenGL                 
[I]               main.c:839  | run                            | Using: OpenGL                          
[I]               main.c:976  | run                            | Waiting for host to signal it's ready..
.                                                                                                       
[I]             opengl.c:455  | opengl_render_startup          | Vendor  : X.Org                        
[I]             opengl.c:456  | opengl_render_startup          | Renderer: AMD Radeon HD 8500 series (OL
AND, DRM 3.26.0, 4.18.0-16.1-liquorix-amd64, LLVM 6.0.1)                                                
[I]             opengl.c:457  | opengl_render_startup          | Version : 3.1 Mesa 18.1.9              
[I]             opengl.c:469  | opengl_render_startup          | Using GL_AMD_pinned_memory             
[I]               main.c:985  | run                            | Host ready, starting session           
[I]             opengl.c:890  | configure                      | Using decoder: NULL                    
[E]             opengl.c:859  | _check_gl_error                | 935: glGenBuffers = 1281 (invalid value
)                                                                                                       
[I]             opengl.c:890  | configure                      | Using decoder: NULL   

This is only on Wayland, Xorg works.

gnif:
Wayland is not supported at this time. You may have better luck using the newer EGL renderer, to use it run with `looking-glass-client -g egl`

Yardanico:
@gnif Same happens for me, however I don't use Wayland (I use LXQt which is Xorg-only).
Host CPU: i5-4460
Host GPU: Intel HD 4600
Guest GPU: NVIDIA GTX 750
[...]

PR #108, opened 2018-12-16
Store the initial window size in the state

ananace:
The SDL resize event only gets fired when the window is resized from it's starting size, so if the user doesn't resize the window, then no size will ever be recorded into the state.

Without a size in place there both the OpenGL and the EGL renderer seem to break, though in different ways - OpenGL crashes with a glGenBuffer error (#101 #107), and EGL simply renders a black screen.

This reads the window size initially as well as on resizes, and fixes said issues at least on my system.

### DIFF
--- a/client/main.c
+++ b/client/main.c
@@ -862,6 +862,9 @@ int run()
   if (!params.center)
     SDL_SetWindowPosition(state.window, params.x, params.y);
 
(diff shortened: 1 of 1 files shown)


gnif:
Nice catch, I have been indenting to fix this but it kept slipping my mind.

PR #125, opened 2019-01-20
Fix: remove obsolete spice error codes header

the-eater:
Header was removed 4 weeks ago: https://gitlab.freedesktop.org/spice/spice-protocol/commit/334cef51d03134a50d85827ea3c2ebc1ebb09e1c

### DIFF
--- a/client/spice/spice.c
+++ b/client/spice/spice.c
@@ -36,7 +36,6 @@ Place, Suite 330, Boston, MA 02111-1307 USA
 #include <arpa/inet.h>
 
 #include <spice/protocol.h>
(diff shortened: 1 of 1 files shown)


the-eater:
While I do understand your point, and agree with you that the protocol errors should be part of the headers. It does not fix the current issue.

Since `0` is pretty much the standard "nothing went wrong" return and we don't use any other error codes currently, I would suggest removing inclusion of this header for now.

Currently LookingGlass will _not_ build against the latest version of spice-protocol and since I'm expecting most users will be using a rolling release/bleeding edge issue, this will become a bigger issue.

PR #127, opened 2019-01-24
change spice port type to unsigned short

flumm:
so that ports >32767 get displayed correctly,
also signed overflow is undefined behaviour

Signed-off-by: Dominik Csapak <d.csapak@proxmox.com>


gnif:
Thanks!

PR #134, opened 2019-02-20
[client] update viewport size after window is created

detergnet:
Make sure glViewport gets called as soon as we know the window
dimensions, otherwise nothing gets rendered until KVMFR communication is
established (mostly the splash).

Without the fix, './looking-glass-client -s' behaves differently than
'./looking-glass-client -s -F' in that the latter triggers a
SDL_WINDOWEVENT_SIZE_CHANGED and updates the viewport immediately after
window creation while the former doesn't and all rendering is delayed
until the frameThread successfully decodes the first frame.

As all the possible ways of updating the viewport (window creation,
window size change, frame size change) are covered with
updatePositionInfo(), the 'started' state becomes redundant and is
removed.

Note: this might be the wrong way to fix it (possible driver bug?),
glViewport's specification mentions that the default viewport size
matches the window size when the GL context is first attached.

Tested on:

Debian Buster with GNOME 3.30.2 on X.Org 1.20.3
Looking Glass (a12-21-g07e4c1c20f)
Locking Method: Atomic
Using: EGL
Vendor  : Intel Open Source Technology Center
Renderer: Mesa DRI Intel(R) Haswell Desktop
Version : OpenGL ES 3.1 Mesa 18.3.2

Signed-off-by: Marius Barbu <msb@avengis.com>

### DIFF
--- a/client/main.c
+++ b/client/main.c
@@ -46,7 +46,6 @@ Place, Suite 330, Boston, MA 02111-1307 USA
 struct AppState
 {
   bool                 running;
-  bool                 started;
   bool                 keyDown[SDL_NUM_SCANCODES];
 
   bool                 haveSrcSize;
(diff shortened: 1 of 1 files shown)


gnif:
Looks good, thanks!

PR #157, opened 2019-05-09
Added support for IVSHMEM device selection on old `host` and new `c-host` for windows

paul70078:
Device selection on `c-host` is very simple at the moment. It would be possible to extend it to specify ivshmem devices not only by member index, but also by pci bus location (bus, slot and optional function)


gnif:
Look good but please squash these commits into a single commit as there is no need to have all this revision history. Also please be sure to re-base your changes on top of the master branch, general rule of thumb, merging is bad, always re-base.

Thanks

paul70078:
I've made the changes :)

gnif:
Thanks for your efforts in this.

PR #164, opened 2019-05-26
[client] Added option to maximize on startup

JJRcop:
Added `win:maximize` or `-T` option to the client which maximizes the client on startup.
It's off by default


gnif:
Brilliant, thanks!

PR #167, opened 2019-06-04
KVMFR Module build file update

Omar007:
Update to allow the module to be build for different kernels. Needed for DKMS.
Also updated the DKMS file to make use of that.

I've not made other changes to the DKMS file  (such as removing the REMAKE_INITRD and changing the module location) as I do not know what system the file was originally targeting and if that actually does need those.

### DIFF
--- a/module/Makefile
+++ b/module/Makefile
@@ -1,11 +1,13 @@
 obj-m += kvmfr.o
 USER  := $(shell whoami)
+KVER ?= $(shell uname -r)
+KDIR ?= /lib/modules/$(KVER)/build
 
 all:
-	make -C /lib/modules/$(shell uname -r)/build/ M=$(PWD) modules
+	make -C $(KDIR) M=$(PWD) modules
 
 clean:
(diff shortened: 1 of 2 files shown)


gnif:
I am not entirely familiar with the DKMS configuration and simply altered a template from the DKMS man page, if `REMAKE_INITRD` is not required please feel free to remove it.

Omar007:
Ok I'll make a rewrite of the dkms.conf as well and squash that one commit.

Omar007:
I've also changed the group of the module so that any/all modules that would ever exist as part of this project would be grouped together when installed through dkms. (pre-commit updated for that reason)
If you're not ok with let me know and I'll revert that to package/group==module and just correct the package version to match the version specified in the module itself.

Omar007:
Appareantly still half asleep as obv. that pre-commit change would not have worked the next time it was run >.>

gnif:
I don't agree with the rename of the group, it is KVMFR, not Looking Glass. Looking Glass implements the KVMFR protocol.

PR #176, opened 2019-06-30
Add missing linker libraries

Michal-Szczepaniak:
On opensuse tumbleweed those missing libs are stopping lookingglass from linking.


gnif:
I am sorry but there are several issues with this patch

1) It's platform specific to OpenSuse, you need to detect the platform and only add these for it.
2) Wrong file, please see `common/CMakeLists.txt` line 35.

gnif:
Closing this PR since it has seen no activity, I will re-open it when/if these issues are addressed.

PR #178, opened 2019-07-04
[c-host] nvfbc: support relative path to SDK

jonpas:
Adds support to define `NVFBC_SDK` path as relative:
```bash
$ cmake -DCMAKE_TOOLCHAIN_FILE=../toolchain-mingw64.cmake -DNVFBC_SDK='../NVIDIA Capture SDK' -DUSE_NVFBC=1 ..
```

Really just a small quality of life thing when having SDK located in some deep folder.

### DIFF
--- a/c-host/platform/Windows/capture/NVFBC/CMakeLists.txt
+++ b/c-host/platform/Windows/capture/NVFBC/CMakeLists.txt
@@ -6,6 +6,10 @@ add_library(capture_NVFBC STATIC
 	src/wrapper.cpp
 )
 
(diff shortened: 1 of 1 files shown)


gnif:
While I appreciate the idea this will not work when building in an directory outside of the git tree which is often done.

jonpas:
I see. Would it not be better to at least support this partially at least compared to not at all?

gnif:
no, because it will cause confusion as documentation will be inaccurate for those trying to build out of tree.

PR #209, opened 2019-11-20
[client] Use eglGetPlatformDisplay() to fix surface creation

aqxa1:
[Why]
Recent versions of Mesa may have trouble with surface creation, resulting in
errors like:
egl.c:428  | egl_render_startup             | Failed to create EGL surface (eglError: 0x300b)

[How]
Replace eglGetDisplay() with eglGetPlatformDisplay(). Requires EGL 1.5, but should
be supported with any desktop driver released in the past few years.

I haven't tested with Wayland but it should be correct.

### DIFF
--- a/client/renderers/EGL/egl.c
+++ b/client/renderers/EGL/egl.c
@@ -53,7 +53,6 @@ struct Inst
   LG_RendererParams params;
   struct Options    opt;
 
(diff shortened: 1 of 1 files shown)


gnif:
Thanks for this, I will test this out when I return home (next week).

PR #210, opened 2019-11-25
[client] do not force disable compositor

thejavascriptman:
This PR avoids force disabling the compositor so the `SDL_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR` environment variable gets honored.

By default SDL disables compositor so the default behavior will not be changed, however if the user sets `SDL_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR=0`, SDL will not disable compositor which is necessary to avoid tearing when running under kwin.

I didn't add an additional command line argument to avoid cluttering and `SDL_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR` is supported out of the box by SDL.

Please let me know if it looks good to you, if there is a specific compositor that does not get disabled by SDL I would like to add an additional check rather than disabling compositor completely e.g.
```
const char* bypass = getenv("SDL_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR");
SDL_SetHint(SDL_HINT_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR, bypass == NULL || strcmp(bypass, "1") == 0 ? "1" : "0");
```

### DIFF
--- a/client/src/main.c
+++ b/client/src/main.c
@@ -1192,30 +1192,12 @@ static int lg_run()
 
   register_key_binds();
 
-  // set the compositor hint to bypass for low latency
   SDL_SysWMinfo wminfo;
   SDL_VERSION(&wminfo.version);
   if (SDL_GetWindowWMInfo(state.window, &wminfo))
   {
     if (wminfo.subsystem == SDL_SYSWM_X11)
     {
-      Atom NETWM_BYPASS_COMPOSITOR = XInternAtom(
-        wminfo.info.x11.display,
-        "NETWM_BYPASS_COMPOSITOR",
-        False);
-
-      unsigned long value = 1;
-      XChangeProperty(
-        wminfo.info.x11.display,
-        wminfo.info.x11.window,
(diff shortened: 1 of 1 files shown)


gnif:
Instead of changing the default behavior of the client which most people now expect this would be better suited to be an application option that defaults to bypass. Please see https://github.com/gnif/LookingGlass/blob/master/client/src/config.c as to how to add an option.

thejavascriptman:
The default behavior will be preserved as SDL disables compositor by default (please see https://discourse.libsdl.org/t/sdl-2-0-8-released/23957)

Quote from the link above:
> Added the hint SDL_HINT_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR to control whether the X server should skip the compositor for the SDL application. This defaults to “1”

Is there an specific scenario on which the default behavior could change?

gnif:
Some builds do not on various distributions, it's why this was added in the first place.

thejavascriptman:
Thank you for the explanation. I'm testing a different approach, will make a new PR soon.

PR #212, opened 2019-11-29
Optionally reenable compositor when client window loses focus

thejavascriptman:
This PR adds the optional `win:enableCompositorOnFocusLoss` argument to re-enable compositor if client window loses focus. 

The default value for `enableCompositorOnFocusLoss` is `false` so default behavior should be preserved, however I had to replace `NETWM_BYPASS_COMPOSITOR` with `_NET_WM_BYPASS_COMPOSITOR` (the first one was not working at all).

Please let me know if there is anything else that can be improved or if you have a better name for the new parameter 🙂


client/src/main.c
```diff
@@ -583,6 +583,30 @@ void spiceClipboardRequest(const SpiceDataType type)
     state.lgc->request(spice_type_to_clipboard_type(type));
 }
 
+// set the compositor hint to bypass for low latency
+void bypassCompositor(unsigned long value)
+{
+  if (state.wminfo.subsystem != SDL_SYSWM_X11)
```
gnif: cosmetic: no need for curly braces here, please change this to the format
```
if (boolean)
  return;
```

client/src/main.c
```diff
@@ -583,6 +583,30 @@ void spiceClipboardRequest(const SpiceDataType type)
     state.lgc->request(spice_type_to_clipboard_type(type));
 }
 
+// set the compositor hint to bypass for low latency
+void bypassCompositor(unsigned long value)
```
gnif: this should be `static void` as it's only used locally, and please use a boolean, not unsigned long.

client/src/main.c
```diff
@@ -618,6 +642,17 @@ int eventFilter(void * userdata, SDL_Event * event)
         case SDL_WINDOWEVENT_CLOSE:
           state.running = false;
           break;
+
+        case SDL_WINDOWEVENT_FOCUS_LOST:
+          if (params.enableCompositorOnFocusLoss)
```
gnif: cosmetic: ditto re braces

client/src/main.c
```diff
@@ -1192,41 +1227,21 @@ static int lg_run()
...
-        1
-      );
-
-      state.lgc = LG_Clipboards[0];
-    }
-  } else {
     DEBUG_ERROR("Could not get SDL window information %s", SDL_GetError());
     return -1;
   }
+  if (state.wminfo.subsystem == SDL_SYSWM_X11)
```
gnif: new line before this `if` please and again no need for the braces

thejavascriptman:
Thank you for the detailed feedback.

Regarding the parameter name, do you recommend implementing an enum like
```
enum AllowCompositor
{
  NO=0,
  YES=1,
  ON_FOCUS_LOST=2
}
```

Or just use an integer and mention it's values in the readme file? e.g.
```
| win:allowCompositor |    |  0  | Allow compositor (0=no,  1=yes, 2=OnLostFocus)  |
```

thejavascriptman:
Thank you for the comments, will take a look during this week

thejavascriptman:
@gnif Please take a look and let me know if there are other changes you would like me to make.

gnif (changes requested):
Sorry but this PR is now a complete jumble, seems you made a mistake when merging. I suggest you rebase your changes on the latest master and void merges in the future which will avoid these issues.

gnif:
As the code base has changed so much this PR doesn't make any sense anymore as it stands. Closing.

PR #219, opened 2019-12-17
[client] fix return value in spice_read_nl error path

rikardfalkeborn:
Returning -1 from a function with bool as return argument is the same as
returning true. If the channel is not connected, return false instead to
indicate the error.


gnif:
Well spotted, thanks!

PR #234, opened 2020-01-27
Update README.md

benedikt-bartscher:
Clone recursive to include LGMP


gnif:
Git should not be included in the build instructions as most people are building from a complete tarball.

Issue #232, opened 2020-01-14
binutils version 2.33 renames bfd_get_section_flag macro


lhw:
With commit [fd361982](https://sourceware.org/git/gitweb.cgi?p=binutils-gdb.git;h=fd3619828e94a24a92cddec42cbc0ab33352eeb4) of binutils some of the macros were renamed, including bfd_get_section_flag. This was reported on the [debian bugtracker](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=948529). My simple fix was to switch the macro for now but this would of course break compatibility with older versions of binutils. Concerning the other two points he brought up thats more of a debian workaround.

chrsm:
This got it to build and run for me, though I'm not sure if it's 100% correct :-)

```c
diff --git a/common/src/platform/linux/crash.c b/common/src/platform/linux/crash.c
index 218d65f..7b08a41 100644
--- a/common/src/platform/linux/crash.c
+++ b/common/src/platform/linux/crash.c
@@ -98,10 +98,10 @@ static void load_symbols()

 static bool lookup_address(bfd_vma pc, const char ** filename, const char ** function, unsigned int * line, unsigned int * discriminator)
 {
-  if ((bfd_get_section_flags(crash.fd, crash.section) & SEC_ALLOC) == 0)
+  if ((bfd_section_flags(crash.section) & SEC_ALLOC) == 0)
     return false;

-  bfd_size_type size = bfd_get_section_size(crash.section);
+  bfd_size_type size = bfd_section_size(crash.section);
   if (pc >= size)
     return false;

@@ -230,4 +230,4 @@ bool installCrashHandler(const char * exe)
   return true;
 }

-#endif
\ No newline at end of file
+#endif
```

gnif:
The change is simple but for now, it's on hold as only those running on unstable/bleeding edge versions of Binutils are affected. Really this should have some `ifdefs` to select the correct code depending on the binutils version.

chrsm:
@gnif Sorry for the delay. I made a PR with these changes properly ifdef'd.

https://github.com/gnif/LookingGlass/pull/241

gnif:
Looks and tests good, thanks!

Fixed in #241

PR #241, opened 2020-02-22
[common] fix build for newer versions of binutils

chrsm:
binutils has changed several macros. Added ifdef to allow building with
stable and bleeding edge versions.

refs #232


gnif:
Awesome, thanks for this, I had intended to address it but had not yet got around to it.

PR #247, opened 2020-03-07
Update README.md

stefanleh:
Add instructions for Fedora linux.


gnif:
The readme is pending removal in favor of the wiki for such information, as such updates to this file for other platforms just adds noise to the commit history.

PR #258, opened 2020-03-28
fix a buffer overflow in opengl options

fisherwise:


gnif:
Thanks for this, could you, however, please squash these commits into a single commit.

fisherwise:
I created another PR. Closing this one.

PR #259, opened 2020-03-30
Tip for git submodules

wheelcomplex:


gnif:
If you are using git to begin with and building from the source there should be no need to handhold.

PR #270, opened 2020-04-23
[client] add quit keybind

feltcat:
Adds a <kbd>ScrLk</kbd>+<kbd>Q</kbd> keybind to conveniently quit the client.

On a side note, I noticed that all the keybinds are in `release_key_binds()` except for the mouse sensitivity ones. Is this intentional, or an oversight?


gnif:
Good addition thanks, I will merge this.

> except for the mouse sensitivity ones. Is this intentional, or an oversight?

Oversight... I will fix it :)

PR #278, opened 2020-05-21
[client] add option to capture input on start

SamuelBowman:
Adds the option spice:captureOnStart which automatically captures keyboard and mouse input when Looking Glass starts. Defaults to false in order to maintain current default behavior.

This allows for the creation of a single keyboard shortcut which launches Looking Glass in full screen mode and automatically captures input (equivalent functionality to pressing the button on a physical KVM switch). Previously a shortcut could be created to launch Looking Glass, but you still had to press scroll lock (or whatever the escape key has been mapped to) to begin capturing input.

### DIFF
--- a/client/src/config.c
+++ b/client/src/config.c
@@ -309,6 +309,13 @@ static struct Option options[] =
     .type           = OPTION_TYPE_BOOL,
     .value.x_bool   = true
   },
+  {
+    .module         = "spice",
+    .name           = "captureOnStart",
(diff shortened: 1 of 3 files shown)


JJRcop:
While a seemingly strange workflow, this PR enables you to set up your DE to open looking glass in this manner and press `ScrLk` **+** `Q` to exit looking glass, you could do this multiple times very quickly.

PR #280, opened 2020-05-21
[Host] Add host build instructions for WSL

TheCakeIsNaOH:


TheCakeIsNaOH:
Rebased on the latest master after c-host was renamed to host.

gnif:
Thanks for rebasing... I am holding off on this for the moment as your work has inspired me to figure out if we can launch LG with SYSTEM level access for UAC capture etc. I cracked it last night and have a proof of concept working. It will be a system service that stats LG. Once it's polished and in the repo I will revisit this so it can be incorporated directly into the installer.

TheCakeIsNaOH:
Awesome, I'll just check back later then.

gnif:
I am sorry but looking at this I question if it's even really needed, it's basically just saying to install WSL before the normal build process.

PR #291, opened 2020-07-19
Add option to skip splash screen fade out

JJRcop:
Inspired by https://github.com/gnif/LookingGlass/pull/278#issuecomment-659531054, this option removes the fade out when a looking glass connection is established.

It will help if someone starts and quits the client often.


JJRcop:
VERSION was wack because I amended the commit and the hook ran again

gnif:
As we are in a release candidate phase new features are not being introduced, only bug fixes. This will be postponed until B2 is out.

JJRcop:
I still don't mind waiting for B3, but I think you've cheated a little bit on the "only bug fixes" stipulation. :laughing:

gnif:
I disagree, all changes that have entered are to address bugs/issues that have been raised. Your PR here is adding something completely new that only you have requested.

ie: 
* Running as a service allows us to fix a long-standing performance bug with thread priority as it gives us system elevation. It also allows us to finally use an NSIS installer to make B2 (when released) far easier to install, which is another long-standing issue.
* Cursor hotspot and recent mouse alignment changes are also very long-standing bugs.
* Changes that affect latency are based on expected feedback from tagging RC builds and is the entire point.

Please look through the commit history and you will see that pretty much every commit is to adjust/fix some issue.

JJRcop:
You have convinced me. Thanks again for all of your hard work.

gnif:
Merged in 66df00cee2dc0acb3cb71919ac4cf32b4112c3f3

PR #301, opened 2020-08-18
Add spice:captureOnStart to client readme

ash-hat:
Option is from #278 and already merged, but the PR didn't update the readme for the client.


gnif:
Thanks

PR #327, opened 2020-12-01
[resource] basic desktop file

ixevix:


gnif (changes requested):
This is not as simple as renaming the icon and adding this file, the executable can be installed anywhere on the system depending on the install prefix. The CMake project needs to be set up to create and install the desktop file with the path it's configured with, and only for Linux. Renaming the icon for this is also incorrect as the CMake project should do this during installation as it's OS/WM dependent.

ixevix:
I see. I have no experience in CMake so I guess I'll close this.

PR #334, opened 2020-12-08
Replace git diff logic in version.cmake with --dirty

JJRcop:
`git describe` has a `--dirty` argument which does the same thing as the git diff logic, while also including the case of a populated index, which the existing logic missed.

Providing '+' as the argument replicates old behavior.


JJRcop:
I left the first commit in in case you prefer that approach, but I think `--dirty` is better forwards compatible since it is maintained by git developers.

You can squash merge to get rid of the first commit or I can just remove it.

PR #340, opened 2020-12-13
grab/ungrab the keyboard when the mouse enters/exits the window

guruthree:
Cal keyboardGrab() and keyboardUngrab() in the handleWindowEnter() and handleWindowLeave() functions respectively to ensure that the keyboard is only grabbed when the mouse is within the window.

### DIFF
--- a/client/src/main.c
+++ b/client/src/main.c
@@ -956,37 +956,6 @@ static void handleResizeEvent(unsigned int w, unsigned int h)
   updatePositionInfo();
 }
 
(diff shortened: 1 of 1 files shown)


gnif:
Sorry but this causes issues with WMs like i3 where focus follows the cursor. There is a larger issue here I am working on that seems to be triggered by the usage of KDE. When I find some time over the next few days I will investigate, in the interim please revert back to B2.

PR #346, opened 2021-01-04
[doc] fix wiki link in README.md

Xyene:
The previous link was missing a trailing slash, and resolved to a 404
page.


gnif:
the trailing slash is valid, I will fix the server configuration instead.

PR #347, opened 2021-01-04
Fix hang in eglSwapBuffers when exiting while not visible

Xyene:
eglSwapBuffers is allowed to block when called with a nonzero interval
parameter. On Wayland, Mesa will block until a frame callback arrives.
If an application is not visible, a compositor is free to not schedule
frame callbacks (in order to save CPU time rendering something that is
entirely invisible).

Currently, starting Looking Glass from a terminal, hiding it
entirely, and sending ^C will cause Looking Glass to hang joining the
render thread until the window is made visible again.

Calling eglDestroySurface is insufficient to unblock eglSwapBuffers, as
it attempts to grab the same underlying mutex.

Instead, this PR makes it so that we pass a 0 interval to
eglSwapBuffers when running on Wayland, such that we don't block waiting
for a frame callback. This is not entirely ideal as it *does* mean
Looking Glass submits buffers while hidden, but it seems better than
hanging on exit.

It also removes the opengl:vsync and egl:vsync flags when running on
Wayland, as they are meaningless there.

### DIFF
--- a/client/renderers/EGL/egl.c
+++ b/client/renderers/EGL/egl.c
@@ -97,6 +97,16 @@ struct Inst
   LG_FontObj        fontObj;
 };
 
+static bool egl_vsync_option_validator(struct Option * opt, const char ** error)
+{
+  if (opt->value.x_bool && getenv("WAYLAND_DISPLAY"))
+  {
+    DEBUG_WARN("Cannot disable vsync on Wayland, forcing egl:vsync=off");
+    opt->value.x_bool = false;
+  }
+
+  return true;
+}
 
 static struct Option egl_options[] =
 {
@@ -105,7 +115,8 @@ static struct Option egl_options[] =
     .name         = "vsync",
     .description  = "Enable vsync",
     .type         = OPTION_TYPE_BOOL,
-    .value.x_bool = false
+    .value.x_bool = false,
+    .validator    = &vsync_option_validator
   },
   {
     .module       = "egl",
--- a/client/renderers/OpenGL/opengl.c
+++ b/client/renderers/OpenGL/opengl.c
@@ -62,7 +62,7 @@ static struct Option opengl_options[] =
     .name         = "vsync",
     .description  = "Enable vsync",
     .type         = OPTION_TYPE_BOOL,
-    .value.x_bool = true
+    .value.x_bool = false
   },
   {
     .module       = "opengl",
--- a/client/renderers/OpenGL/opengl.c
+++ b/client/renderers/OpenGL/opengl.c
@@ -48,6 +48,17 @@ Place, Suite 330, Boston, MA 02111-1307 USA
 
 #define FADE_TIME 1000000
 
+static bool opengl_vsync_option_validator(struct Option * opt, const char ** error)
+{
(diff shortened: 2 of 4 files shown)


client/renderers/OpenGL/opengl.c
```diff
@@ -192,6 +197,9 @@ const char * opengl_get_name()
 static void opengl_setup()
 {
   option_register(opengl_options);
+  if (!getenv("WAYLAND_DISPLAY")) {
```
Xyene, line 200: I'm happy to move this logic into some sort of `is_wayland` helper, but am not sure where that should live. Perhaps `utils.c`?

client/renderers/EGL/egl.c
```diff
@@ -145,6 +138,18 @@ static struct Option egl_options[] =
...
+  {
+    .module       = "egl",
+    .name         = "vsync",
+    .description  = "Enable vsync",
+    .type         = OPTION_TYPE_BOOL,
+    .value.x_bool = false
+  },
+  {0}
+};
+
```
gnif, line 152: No, keep this part of the standard EGL options. vsync disable is not just `x11` or `Wayland`. It would be better to add a validator that warns/errors if it's enabled on Wayland if it really warrants this.

client/renderers/EGL/egl.c
```diff
@@ -155,6 +160,9 @@ const char * egl_get_name()
 void egl_setup()
 {
   option_register(egl_options);
+  if (!getenv("WAYLAND_DISPLAY")) {
+    option_register(opengl_options_x11);
+  }
```
gnif, line 165: ditto

client/renderers/OpenGL/opengl.c
```diff
@@ -59,22 +59,27 @@ static struct Option opengl_options[] =
...
+};
+
+static struct Option opengl_options_x11[] =
+{
   {
     .module       = "opengl",
-    .name         = "amdPinnedMem",
-    .description  = "Use GL_AMD_pinned_memory if it is available",
+    .name         = "vsync",
+    .description  = "Enable vsync",
```
gnif, line 82: ditto

client/src/main.c
```diff
@@ -1478,12 +1478,7 @@ static int lg_run()
        if (g_cursor.sens < -9) g_cursor.sens = -9;
   else if (g_cursor.sens >  9) g_cursor.sens =  9;
 
-  char* XDG_SESSION_TYPE = getenv("XDG_SESSION_TYPE");
-
-  if (XDG_SESSION_TYPE == NULL)
-    XDG_SESSION_TYPE = "unspecified";
-
-  if (strcmp(XDG_SESSION_TYPE, "wayland") == 0)
+  if (getenv("WAYLAND_DISPLAY"))
```
gnif, line 1481: these changes look fine

gnif (commented):
The fix for the issue is sound, but we should not be confusing people by having disappearing configure options, instead add a validator to the vsync options to check for Wayland and tell the user they can't use it on Wayland.

gnif:
Perfect! Thanks!

PR #348, opened 2021-01-04
[client] spice: properly handle high-precision scroll wheel input

Xyene:
Some setups (e.g. Wayland) have high precision scroll wheel input, such
that the y-delta on an event may exceed 1. In these cases, scrolling up
currently gets treated as scrolling down.

This commit changes the checks to use > 0 rather than == 1.

This is the approach suggested in
https://wiki.libsdl.org/SDL_MouseWheelEvent.

### DIFF
--- a/client/src/main.c
+++ b/client/src/main.c
@@ -1287,8 +1287,8 @@ int eventFilter(void * userdata, SDL_Event * event)
         break;
 
       if (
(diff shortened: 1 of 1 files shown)


gnif:
Looks good, but please adjust the commit message to the format `[client] spice: xxx`

Xyene:
Done :slightly_smiling_face:

PR #349, opened 2021-01-04
[client] opengl: require 8 bit colour when creating context

quantum5:
For some reason, if we don't ask for 8-bit, wayland will give a context with
lower bit depth, which looks terrible.


gnif:
Looks good but please adjust your commit message to `[client] opengl: xxx`

quantum5:
Done.

Issue #306, opened 2020-09-02
Resize events not handled correctly on Wayland-EGL


Xyene:
### Required information

Host CPU: Ryzen 3900X
Host GPU: RX 480
Guest GPU: GTX 1060
Host Kernel version: 5.4.0
Host QEMU version: 5.0.0

It looks like the Wayland EGL renderer is swapping buffers in such a way that resize events aren't triggered, leading to the guest not being resized when running with `-a`.

![image](https://user-images.githubusercontent.com/1403503/91929495-84841480-ecac-11ea-8a54-99aeadc541fa.png)

I believe I've tracked down why this happens.

When SDL receives an `xdg_surface.configure` event, it hits [`handle_configure_xdg_shell_surface`](https://github.com/spurious/SDL-mirror/blob/96f1313583bb7bb03f4cd936d4aee5b2a29066a7/src/video/wayland/SDL_waylandwindow.c#L221).  If it's not the very first configure event, it hits [this](https://github.com/spurious/SDL-mirror/blob/96f1313583bb7bb03f4cd936d4aee5b2a29066a7/src/video/wayland/SDL_waylandwindow.c#L251-L253) bit of code:

```c
        if (!(window->flags & SDL_WINDOW_OPENGL)) {
            Wayland_HandlePendingResize(window);  /* OpenGL windows handle this in SwapWindow */
        }
```

The `SDL_WINDOW_OPENGL` is set for both OpenGL and EGL windows. `SwapWindow` for EGL surfaces is [defined](https://github.com/spurious/SDL-mirror/blob/69b8870c9c9a5e4bee56a6bae63bde4113749735/src/video/wayland/SDL_waylandopengles.c#L62-L77) like:

```c
int
Wayland_GLES_SwapWindow(_THIS, SDL_Window *window)
{
    SDL_WindowData *data = (SDL_WindowData *) window->driverdata;

    if (SDL_EGL_SwapBuffers(_this, data->egl_surface) < 0) {
        return -1;
    }

    // Wayland-EGL forbids drawing calls in-between SwapBuffers and wl_egl_window_resize
    Wayland_HandlePendingResize(window);

    WAYLAND_wl_display_flush( data->waylandData->display );

    return 0;
}
```

`Wayland_HandlePendingResize` is in charge of firing off `SDL_WINDOWEVENT_RESIZED`.

[...]

PR #350, opened 2021-01-04
[client] egl: disable EGL when running on Wayland

Xyene:
This commit makes Looking Glass always use the OpenGL renderer when
running on Wayland. The EGL renderer is broken on Wayland and can't
reasonably be fixed until SDL is dropped entirely (as per
https://github.com/gnif/LookingGlass/issues/306).

Until that time, the OpenGL renderer provides a much better
Wayland-native experience.

### DIFF
--- a/client/renderers/EGL/egl.c
+++ b/client/renderers/EGL/egl.c
@@ -159,6 +159,13 @@ void egl_setup()
 
 bool egl_create(void ** opaque, const LG_RendererParams params)
 {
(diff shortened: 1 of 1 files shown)


Xyene:
For comparison, Wayland-OpenGL:

image

Wayland-EGL:

image

gnif:
Looks good, but please adjust your commit message to the following format `[client] egl: xxx`

Xyene:
Done :+1:

PR #351, opened 2021-01-04
[host/client] scale mouse input based on host DPI information

quantum5:
When running Windows with UI scaling, Windows also scales the speed of mouse movement. This causes the mouse movements to get out of sync. For example, running with `-M` with 125% UI scaling shows the Windows cursor moving 125% faster than the host machine cursor.

This PR fetches the UI scaling information for the captured monitor and passes it along inside `KVMFRFrame`. This is then used to scale the mouse movements sent to spice.

To obtain UI scaling information, the host is unconditionally made per-monitor DPI aware, and not just in the DXGI path. As a bonus, this also fixes an NvFBC bug with scaling that results in the image being cropped at the virtual resolution reported to non-DPI aware apps.

### DIFF
--- a/client/src/main.c
+++ b/client/src/main.c
@@ -154,10 +154,12 @@ static void updatePositionInfo()
 
     g_cursor.scale = (
         g_state.srcSize.y != g_state.dstRect.h ||
-        g_state.srcSize.x != g_state.dstRect.w);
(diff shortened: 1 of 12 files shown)


gnif:
Commited in a172d79f6641135f8c2622a95dc2c9bd4e15b34f

Issue #138, opened 2019-03-10
Feature suggestion: -Q should ignore only some quit requests, not all


trying-lg1:
Host CPU: AMD Ryzen 7 2700X
Host GPU: MSI Nvidia 1030 2GB GDDR5, GP108
Guest GPU: Sapphire AMD Radeon R9 380X 4GB GDDR5
Host Kernel version: 4.19, manjaro/arch
Host QEMU version: 3.1.0.1

LG client suggestion:
-Q argument shouldn't ignore if I try to quit the LG client from its menu (or from Gnome overview via X). = When I deliberately switch focus outside guest/client and want to close it.

PR #352, opened 2021-01-05
[client] fix ignoreQuit on Wayland

quantum5:
On Wayland, SDL_WINDOWEVENT_CLOSE is sent even when exiting with keyboard
shortcuts. This meant that the client is still closed even with -Q.

We now swallow SDL_WINDOWEVENT_CLOSE if the cursor is inside the VM. This
should prevent keyboard shortcuts from closing the client, while still
allowing the window to be closed by clicking X with the mouse per #138.

### DIFF
--- a/client/src/main.c
+++ b/client/src/main.c
@@ -1099,9 +1099,9 @@ int eventFilter(void * userdata, SDL_Event * event)
           }
           break;
 
-        // allow a window close event to close the application even if ignoreQuit is set
         case SDL_WINDOWEVENT_CLOSE:
-          g_state.state = APP_STATE_SHUTDOWN;
+          if (!params.ignoreQuit || !g_cursor.inView)
+            g_state.state = APP_STATE_SHUTDOWN;
           break;
       }
       return 0;


client/src/main.c
```diff
@@ -1099,9 +1099,14 @@ int eventFilter(void * userdata, SDL_Event * event)
           }
           break;
 
-        // allow a window close event to close the application even if ignoreQuit is set
+        // Allow a window close event to close the application even if ignoreQuit is set,
+        // but only when the cursor is not in the client area.
+        // On Wayland, keyboard shortcuts to kill windows trigger a close event, so they
+        // should be ignored if triggered while the mouse is in the VM.
```
gnif, line 1105: No need for the verbose comment, one can always look to git blame to see why this code was added. Also this should be the same for xorg anyway too :)

gnif (changes requested):

client/src/main.c
quantum5: I removed the whole comment since how the handler affects `ignoreQuit` should be clear from the code.

PR #355, opened 2021-01-09
Fix various bugs related to the mouse on Wayland

quantum5:


client/src/main.c
```diff
@@ -1126,6 +1131,7 @@ static void handleWindowLeave()
 static void handleWindowEnter()
 {
   g_cursor.inWindow = true;
+  g_state.focused   = true;
```
gnif, line 1134: Entered is not the same as focused, this is incorrect. Enter means the mouse is over the window, but the window may not have input focus.

quantum5: Fixed.

gnif:
I think you're trying to fix here a bug we are seeing on X11 also. The trigger is that if the mouse is for any reason moved outside of the window such as toggling full screen while the cursor is in the upper right quadrant of the LG window, the application can not be re-focused and as such clicking breaks.

Also, how will Wayland work if you can't grab the cursor anymore? All movements are based on relative x/y and if the cursor moves outside of the LG client, input will stop.

gnif:
Please see 76ed75f871847111714bc271afb5f780cde45740. This should help with this also and for the most part, make this PR obsolete.

quantum5:
Superseded by #359.

PR #359, opened 2021-01-10
Add Wayland-specific mouse logic due to lack of warp

quantum5:
Use different logic on Wayland to avoid warp. This requires a 1:1 mouse movement patch in the guest (when using Wayland), but it makes the desktop usable. For gaming, capture mode is recommended.

### DIFF
--- a/client/src/main.c
+++ b/client/src/main.c
@@ -913,6 +913,25 @@ static void guestCurToLocal(struct DoublePoint *local)
   local->y = (g_cursor.guest.y + g_cursor.guest.hy) / g_cursor.scale.y;
 }
 
+// On Wayland, our normal cursor logic does not work due to the lack of cursor
+// warp support. Instead, we attempt a best-effort emulation which works with a
+// 1:1 mouse movement patch applied in the guest. For anything fancy, use
+// capture mode.
+static void handleMouseWayland()
+{
+  double ex = (g_cursor.pos.x - g_cursor.guest.x) / g_cursor.dpiScale;
+  double ey = (g_cursor.pos.y - g_cursor.guest.y) / g_cursor.dpiScale;
+
+  int x, y;
+  cursorToInt(ex, ey, &x, &y);
+
+  g_cursor.delta.x += x;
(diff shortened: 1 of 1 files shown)


client/src/main.c
```diff
@@ -924,6 +924,21 @@ static void guestCurToLocal(struct DoublePoint *local)
   local->y = (g_cursor.guest.y + g_cursor.guest.hy) / g_cursor.scale.y;
 }
 
+// On Wayland, our normal cursor logic does not work due to the lack of cursor
+// warp support. Instead, we attempt a best-effort emulation which works with a
+// 1:1 mouse movement patch applied in the guest. For anything fancy, use
+// capture mode.
+static void handleMouseWayland() {
```
gnif, line 931: nit: `{` on the next line please

client/src/main.c
```diff
@@ -934,6 +949,11 @@ static void handleMouseNormal(double ex, double ey)
     return;
   }
 
+  if (g_state.wminfo.subsystem == SDL_SYSWM_WAYLAND) {
```
gnif, line 952: ditto

client/src/main.c
```diff
@@ -1208,15 +1228,29 @@ int eventFilter(void * userdata, SDL_Event * event)
       switch(event->window.event)
       {
         case SDL_WINDOWEVENT_ENTER:
+          if (g_state.wminfo.subsystem == SDL_SYSWM_WAYLAND)
```
gnif, line 1231: Why is this not just part of the != X11 codepath?

gnif (changes requested):
Few minor things, but looks good

client/src/main.c
quantum5: This is Wayland-specific logic to set `g_cursor.inView`. Normally this is done by `handleMouseNormal`. For the Wayland-specific logic, this needs to be handled by `SDL_WINDOWEVENT_ENTER` and `SDL_WINDOWEVENT_EXIT`.

PR #364, opened 2021-01-12
Implement relative mouse mode for Wayland

Xyene:
This patch series removes SDL from the grabbed mouse handling path on Wayland.

Some build infra changes are necessary to build unstable Wayland protocols `relative-pointer-unstable-v1` and `pointer-constraints-unstable-v1-client-protocol`. These protocols have ~universal adoption, and are what SDL uses internally too.

I've attempted to cleanly separate the Wayland bits through an opaque pointer, in hopes that this will make it easy to move into ADL later.

This PR likely still contains bugs, but I'm dogfooding it for now.

### DIFF
--- a/client/src/main.c
+++ b/client/src/main.c
@@ -88,7 +88,7 @@ struct AppParams params = { 0 };
 static void setGrab(bool enable);
 static void setGrabQuiet(bool enable);
 
-static void handleMouseGrabbed(double ex, double ey);
+void handleMouseGrabbed(double ex, double ey);
 static void handleMouseNormal(double ex, double ey);
 
 static void lgInit()
@@ -883,7 +883,7 @@ static void cursorToInt(double ex, double ey, int *x, int *y)
   *y = (int)ey;
 }
 
-static void handleMouseGrabbed(double ex, double ey)
+void handleMouseGrabbed(double ex, double ey)
 {
   int x, y;
 
@@ -1469,7 +1469,12 @@ int eventFilter(void * userdata, SDL_Event * event)
       g_cursor.pos.y = event->motion.y;
 
       if (g_cursor.grab)
-        handleMouseGrabbed(event->motion.xrel, event->motion.yrel);
+      {
+        // On Wayland, wm.c calls handleMouseGrabbed, bypassing the SDL event
+        // loop.
+        if (g_state.wminfo.subsystem != SDL_SYSWM_WAYLAND)
+          handleMouseGrabbed(event->motion.xrel, event->motion.yrel);
+      }
       else
         handleMouseNormal(event->motion.xrel, event->motion.yrel);
       break;
@@ -2046,6 +2051,8 @@ static int lg_run()
   // the end of the output
   lgWaitEvent(e_startup, TIMEOUT_INFINITE);
 
+  wmInit();
+
   LGMP_STATUS status;
 
   while(g_state.state == APP_STATE_RUNNING)
@@ -2237,7 +2244,10 @@ static void lg_shutdown()
   }
 
   if (g_state.window)
+  {
+    wmFree();
     SDL_DestroyWindow(g_state.window);
+  }
 
   if (cursor)
     SDL_FreeCursor(cursor);
--- a/client/src/wm.c
+++ b/client/src/wm.c
@@ -88,6 +88,9 @@ void wmGrabKeyboard()
         CurrentTime);
       break;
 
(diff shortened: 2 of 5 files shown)


client/src/wm.c
```diff
@@ -155,3 +191,163 @@ void wmWarpMouse(int x, int y)
...
+  dy_unaccel = wl_fixed_to_double(dy_unaccel_w);
+
+  // Add left over fraction from last event.
+  dx_unaccel += wm->dx_frac;
+  dy_unaccel += wm->dy_frac;
+
+  wm->dx_frac = modf(dx_unaccel, &dx);
+  wm->dy_frac = modf(dy_unaccel, &dy);
+
+  fprintf(stderr, "%f %f %d %d\n", dx, dy, dx_unaccel_w, dy_unaccel_w);
```
gnif, line 316: Looks like left behind debug code?

client/src/main.c
```diff
@@ -88,7 +88,7 @@ struct AppParams params = { 0 };
 static void setGrab(bool enable);
 static void setGrabQuiet(bool enable);
 
-static void handleMouseGrabbed(double ex, double ey);
+void handleMouseGrabbed(double ex, double ey);
```
gnif, line 91: While I can understand what you're doing here I am not sure it's the best way to do this. I am OK with this for now, but I may make an agnostic interface for this to make this easier for both X and Wayland code later.

client/src/wm.c
```diff
@@ -23,16 +23,38 @@ Place, Suite 330, Boston, MA 02111-1307 USA
...
+static void wmWaylandInit();
+static void wmWaylandGrabPointer();
+static void wmWaylandUngrabPointer();
+
+void wmInit()
+{
+  switch (g_state.wminfo.subsystem)
+  {
+    case SDL_SYSWM_WAYLAND:
+      wmWaylandInit();
```
gnif, line 52: Missing a `break` after this line

client/src/wm.c
```diff
@@ -155,3 +191,163 @@ void wmWarpMouse(int x, int y)
...
+};
+
+// Registry-handling listeners.
+
+static void registryGlobalHandler(void * data, struct wl_registry * registry,
+    uint32_t name, const char * interface, uint32_t version)
+{
+  struct WMDataWayland * wm = data;
+
+  fprintf(stderr, "%s\n", interface);
```
gnif, line 221: More debug output?
`fprintf` and friends should never be used directly in the LG project, please use `DEBUG_INFO`, `DEBUG_WARN` or `DEBUG_ERROR`.

client/src/wm.c
```diff
@@ -155,3 +191,163 @@ void wmWarpMouse(int x, int y)
...
+{
+  bool hasPointer = capabilities & WL_SEAT_CAPABILITY_POINTER;
+  if (!hasPointer && wm->pointer)
+  {
+    wl_pointer_destroy(wm->pointer);
+    wm->pointer = NULL;
+  }
+  else if (hasPointer && !wm->pointer)
+  {
+    fprintf(stderr, "gained pointer\n");
```
gnif, line 253: ditto

client/src/wm.c
```diff
@@ -155,3 +191,163 @@ void wmWarpMouse(int x, int y)
...
+
+  // Add left over fraction from last event.
+  dx_unaccel += wm->dx_frac;
+  dy_unaccel += wm->dy_frac;
+
+  wm->dx_frac = modf(dx_unaccel, &dx);
+  wm->dy_frac = modf(dy_unaccel, &dy);
+
+  fprintf(stderr, "%f %f %d %d\n", dx, dy, dx_unaccel_w, dy_unaccel_w);
+  handleMouseGrabbed((int)dx, (int)dy);
```
gnif, line 317: Why cast to `int`? `handleMouseGrabbed` takes `double`

client/src/main.c
```diff
@@ -1462,11 +1462,16 @@ int eventFilter(void * userdata, SDL_Event * event)
       if (g_state.wminfo.subsystem == SDL_SYSWM_X11)
         break;
 
-      g_cursor.pos.x = event->motion.x;
-      g_cursor.pos.y = event->motion.y;
+			g_cursor.pos.x = event->motion.x;
+			g_cursor.pos.y = event->motion.y;
```
gnif, line 1466: spaces not tabs please

gnif (changes requested):

PR #353, opened 2021-01-09
[client] add Wayland clipboard support

Xyene:
To do:

- [x] Build scaffolding
- [x] VM-to-host
  - [x] Text selection
  - [x] Image selection
- [x] Host-to-VM
  - [x] Text selection
  - [x] Image selection
- [x] Keyboard hotplug
- [x] Textual data mimetype detection
- [x] ~Multi-seat~ post-ADL only, must have ref to seat that triggered copy operation

PR #365, opened 2021-01-12
[client] wm/wayland: bypass SDL loop for pointer events

Xyene:
Depends on https://github.com/gnif/LookingGlass/pull/353 for conflict reasons, but only the last commit in this chain has changes.


gnif:
branch has conflicts

Xyene:
Rebased on latest masters so that the conflicts are gone.

PR #366, opened 2021-01-12
[client] spice/wayland: correctly update guest cursor location

quantum5:
In the previous commit, the wrong variable was updated.


gnif:
This change is invalid, the delta is used for `autoCapture`.

quantum5:
Yes, `delta` is used for `autoCapture`, which is why it's wrong. This change switches to updating `g_cursor.guest`, which is correct.

gnif:
Ah yes, I missed that this was in the new wayland code. Thanks

PR #378, opened 2021-01-14
[client] spice: correctly use fabs for floating point

quantum5:
The prototype for abs is int abs (int n), which implicitly casts floating
point values to integers. The correct function is fabs.

This commit allows the client to compile under clang.

### DIFF
--- a/client/src/main.c
+++ b/client/src/main.c
@@ -1010,7 +1010,7 @@ void handleMouseNormal(double ex, double ey)
 
   /* if we are in "autoCapture" and the delta was large don't test for exit */
   if (params.autoCapture &&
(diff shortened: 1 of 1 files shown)


client/src/main.c
```diff
@@ -1067,7 +1067,7 @@ void handleMouseNormal(double ex, double ey)
     g_cursor.delta.x += x;
     g_cursor.delta.y += y;
 
-    if (abs(g_cursor.delta.x) > 50 || abs(g_cursor.delta.y) > 50)
+    if (fabs(g_cursor.delta.x) > 50 || fabs(g_cursor.delta.y) > 50)
```
gnif, line 1070: Seeing as this then would be a double, `50` should be `50.0` for correctness.

gnif: Merging as is, I will address this in a future commit.

PR #379, opened 2021-01-14
[client] add support for compiling with MemorySanitizer

quantum5:
This is a sanitizer for tracking access to uninitialized memory.
Violations will be addressed in future commits.


quantum5:
This is not actually that useful as the display driver needs to be instrumented as well to avoid spurious reports.

PR #385, opened 2021-01-15
[host] windows: handle defined exit codes in service

quantum5:
The service should now react to host application exit codes.
For the exit codes that demands it, the service will exit instead of
restarting the host.

### DIFF
--- a/host/platform/Windows/src/service.c
+++ b/host/platform/Windows/src/service.c
@@ -19,6 +19,7 @@ Place, Suite 330, Boston, MA 02111-1307 USA
 
 #include "interface/platform.h"
 #include "common/ivshmem.h"
+#include "service.h"
 
 #include <stdio.h>
 #include <stdbool.h>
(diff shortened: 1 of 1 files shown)


host/platform/Windows/src/service.c
```diff
@@ -692,11 +693,33 @@ VOID WINAPI SvcMain(DWORD dwArgc, LPTSTR *lpszArgv)
...
           doLog("Failed to GetExitCodeProcess (0x%lx)\n", GetLastError());
+        else
+        {
+          doLog("Host application exited with code 0x%lx\n", code);
+          switch (code)
+          {
+            case LG_HOST_EXIT_USER:
+              doLog("Host application exited due to user action\n");
+              goto stopped;
+            case LG_HOST_EXIT_CAPTURE:
```
gnif, line 709: nit: please add whitespace after each break/return/goto

quantum5: Done.

PR #397, opened 2021-01-16
[client] mouse: update cursor draw state even when not in focus

Xyene:
Previously, the cursor would be set to a hidden state when the Looking
Glass window did not have focus, even if the pointer was above the
window, and the window was receiving pointer events (and passed them to
the guest).

### DIFF
--- a/client/src/main.c
+++ b/client/src/main.c
@@ -919,10 +919,16 @@ static void cursorToInt(double ex, double ey, int *x, int *y)
   *y = (int)ey;
 }
 
-static void setCursorInView(bool enable)
+static void setCursorInView(bool inView)
 {
-  // if we don't have focus don't do anything
-  if (enable && !g_state.focused)
+  SDL_ShowCursor(params.hideMouse && inView ? SDL_DISABLE : SDL_ENABLE);
+
+  g_cursor.inView = inView;
+  g_cursor.draw   = params.alwaysShowCursor ? true : inView;
+  g_cursor.redraw = true;
+
+  // if we don't have focus don't try to warp the cursor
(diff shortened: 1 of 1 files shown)


gnif:
This is a UX issue more than anything.

It's correct that the pointer is hidden because people running LG maximized but not fullscreen expect the guest cursor to allow them to exit the client window onto another window that may be partially obscuring the LG window. As the guest and local cursors may not be aligned it is very confusing/jarring when it happens.

Xyene:
> This is a UX issue more than anything.

> As the guest and local cursors may not be aligned it is very confusing/jarring when it happens.

I can see this being the case. I use the 1:1 registry hack and this behavior changed as part of the refactors yesterday.

However, in this case I think the expectation would be that:

- the host cursor isn't hidden when LG doesn't have focus (it currently is, so there's just no cursor displayed over the LG window)
- LG doesn't pass events to the guest when it doesn't have focus (it currently does, so you have on-hover events triggering in the guest despite not seeing the cursor)
- the cursor is shown when the LG window regains focus (it currently isn't)

gnif:
> the host cursor isn't hidden when LG doesn't have focus (it currently is, so there's just no cursor displayed over the LG window)

under Xorg this is not the case, just tested and verified.

> LG doesn't pass events to the guest when it doesn't have focus (it currently does, so you have on-hover events triggering in the guest despite not seeing the cursor)

I just patched this, but yet to push it.

> the cursor is shown when the LG window regains focus (it currently isn't)

Agreed, working on this now.

gnif:
This should all now be addressed in the current master, can you please re-test?

PR #412, opened 2021-01-19
[client] wayland: better handling of keypresses

Xyene:
This PR implements better behavior for when LG loses keyboard focus while keys are held. Previously, any keys held down would remain down, since we did nothing in the leave handler. Now, we send synthetic keyups so the guest isn't left in a confused state.

### DIFF
--- a/client/displayservers/Wayland/wayland.c
+++ b/client/displayservers/Wayland/wayland.c
@@ -245,7 +245,9 @@ static void keyboardEnterHandler(void * data, struct wl_keyboard * keyboard,
 static void keyboardLeaveHandler(void * data, struct wl_keyboard * keyboard,
     uint32_t serial, struct wl_surface * surface)
 {
-  // Do nothing.
+  for (int key = 0; key < KEY_MAX; key++)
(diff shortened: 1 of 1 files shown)


gnif:
If this were to be added it should be done in the main application as it's not Wayland specific. That said, many people including myself take advantage of this in games that require holding a key for long periods, such as running across a large map in a game. It allows one to switch to OBS or a chat application and respond when streaming without interrupting the gameplay.

gnif:
Note, if this is to be a new option it would be a new feature, which would be held until post B3

Xyene:
I was unaware X11 behaved the same in this case, and was considering the lack of leave handling a bug. It's a bit odd -- but usually harmless -- that when executing a window change operation (e.g. <kbd>Mod</kbd> + <kbd>&rarr;</kbd>), in <kbd>Mod</kbd> gets stuck in a pressed state in the VM.

I think there's a case for the release-on-leave behavior since it's what e.g. RDP does, but we can come back to this post-B3.

Xyene:
Superseded by #412.

quantum5:
This is actually superseded by #463.

PR #413, opened 2021-01-20
[client] spice/wayland: fix jitter when moving the cursor slowly

quantum5:
It does not make sense to accumulate fractional error in non-capture mode
as you know exactly where the cursor is supposed to be, at least on Wayland.

On Wayland, we base movements on the current guest position and desired
target position, and the accumulated errors only skew our movements.

### DIFF
--- a/client/src/main.c
+++ b/client/src/main.c
@@ -1438,11 +1438,8 @@ void app_handleMouseBasic()
   struct DoublePoint local;
   guestCurToLocal(&local);
 
-  double lx = (px - local.x) / g_cursor.dpiScale;
-  double ly = (py - local.y) / g_cursor.dpiScale;
-
-  int x, y;
-  cursorToInt(lx, ly, &x, &y);
+  int x = (int) round((px - local.x) / g_cursor.dpiScale);
+  int y = (int) round((py - local.y) / g_cursor.dpiScale);
 
   g_cursor.guest.x += x;
   g_cursor.guest.y += y;


gnif:
What about the factional part caused by the division by `g_cursor.dpiScale`? On X11 we also know the absolute position and in the past when we still were using absolute positional over relative and were rounding as you have done here many people complained.

quantum5:
I think the when we used to absolute positions on X11, we were also looking at the relative movements. I recall the handler taking movement deltas from SDL. Rounding those would create cumulative errors, which we had to correct with the accumulators.

In this implementation for Wayland, mouse movements are completely recalculated every step based on current host and guest cursor positions. Errors do not accumulate (since we always base it on current position), so I think the accumulator is counter-productive.

In practice, I used to observe when moving the mouse horizontally, that there would be small 1px vertical jumps. After this PR, that disappears. As for the fractional part caused by division by `dpiScale`, I am still looking into that, but I don't think this PR makes that part worse.

@Xyene can you confirm if this PR improves tracking on Wayland?

gnif:
I have no issue with this PR, just making sure it's been carefully considered before merging.

Xyene:
The behavior I experience with the existing code is that slow movement is jittery and gets progressively more jittery, and that a sudden fast mouse motion "resets" things for a short period of time. I think there definitely issues with the mouse tracking, but this PR doesn't appear to make them worse in testing, and does improve the small movement case.

Regarding the issues I mentioned, here's some ~straight lines I drew on the Linux host:

image

and my best attempt within the guest:

image

(My mouse definitely did not move _backwards_ at any point.)

PR #414, opened 2021-01-20
[host] nvfbc: force composition to capture some full screen apps

quantum5:
NvFBC is unable to capture certain applications that bypasses the DWM
compositor, for example, Firefox playing video in full screen. This 
has been a known issue for a long time with Nvidia's ShadowPlay, see:
* https://www.nvidia.com/en-us/geforce/forums/geforce-experience/14/233709/                                      
* https://crbug.com/609857

Nvidia won't fix this, but there are workarounds. For example, we
create a transparent 1x1 layered window, which forces desktop composition
to be enabled.

Note that SetLayeredWindowAttributes also supports alpha-based transparency,
but setting transparency to 0 will cause DWM to skip composition. We could 
use a transparency of 1, but this ruins the image by the slightest bit, 
which is unacceptable. Therefore, we must use chroma key-based
transparency, which tricks DWM into compositing despite being fully
transparent.

### DIFF
--- a/host/platform/Windows/include/windows/force_compose.h
+++ b/host/platform/Windows/include/windows/force_compose.h
@@ -1,0 +1,21 @@
+/*
+Looking Glass - KVM FrameRelay (KVMFR) Client
+Copyright (C) 2021 Guanzhong Chen <quantum2048@gmail.com>
+https://looking-glass.io
+
+This program is free software; you can redistribute it and/or modify it under
+the terms of the GNU General Public License as published by the Free Software
+Foundation; either version 2 of the License, or (at your option) any later
+version.
+
+This program is distributed in the hope that it will be useful, but WITHOUT ANY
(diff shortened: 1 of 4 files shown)


host/platform/Windows/include/windows/force_compose.h
```diff
@@ -0,0 +1,21 @@
+/*
+Looking Glass - KVM FrameRelay (KVMFR) Client
+Copyright (C) 2017-2021 Geoffrey McRae <geoff@hostfission.com>
```
gnif, line 3: I am not the author here :)
Please be sure to set correct copyrights where you make alterations/additions, etc.

quantum5: Fixed.

quantum5:
Currently there is a bug that happens when restarting the client.

quantum5:
Bug has been fixed.

PR #419, opened 2021-01-22
[client] spice: keep cursor position when only updating hotspot

quantum5:
g_cursor.guest.x/y stores the coordinates of the top-left corner of the
cursor image. When the cursor message contains CURSOR_FLAG_SHAPE but not
CURSOR_FLAG_POSITION, we update the hotspot location but not the top-left
position. This has the effect of changing the client's view of the guest
cursor position, which is located at (x + hx, y + hy).

In this commit, when we update the cursor shape but not position, we now
keep (x + hx, y + hy) the same. This has the effect of keeping the actual
cursor position the same when the shape changes but not the position.

### DIFF
--- a/client/src/main.c
+++ b/client/src/main.c
@@ -464,6 +464,12 @@ static int cursorThread(void * unused)
           continue;
       }
 
+      if (!(msg.udata & CURSOR_FLAG_POSITION))
+      {
+        g_cursor.guest.x += g_cursor.guest.hx - cursor->hx;
+        g_cursor.guest.y += g_cursor.guest.hy - cursor->hy;
+      }
+
       g_cursor.guest.hx = cursor->hx;
       g_cursor.guest.hy = cursor->hy;
 


gnif:
The guest cursor position should never contain any hotspot component, if it does it is a bug and needs to be fixed in the host application.

quantum5:
The intention is that `g_cursor.guest.x` and `g_cursor.guest.y` always contain the coordinates of the top-left corner of the cursor image, right?

Suppose we have the default Windows pointer at (1000, 500). The hotspot for that cursor is at (0, 0), so `g_cursor.guest.x = 1000`, `g_cursor.guest.y = 500`, `g_cursor.guest.hx = 0`, and `g_cursor.guest.hy = 0`. Now say that cursor changes to the "hourglass" cursor, whose hotspot is at (16, 16), or the centre of the spinner. The point (1000, 500) would be at the centre of the spinner, but the top-left corner of the hourglass cursor would be at (984, 484). We want `g_cursor.guest.x = 984`, `g_cursor.guest.y = 484`, `g_cursor.guest.hx = 16`, and `g_cursor.guest.hy = 16`.

This PR makes this the case when `msg.udata` has `CURSOR_FLAG_SHAPE` but not `CURSOR_FLAG_POSITION`, which is the case because the cursor hasn't actually moved.

gnif:
I get this, however, this PR is working around the fact that the host capture API has not provided an update along with the shape change. This should be fixed in the capture API. I assume this is occurring with NvFBC as DXGI DD provides shape and position updates on such a change.

quantum5:
I see. Will fix it in the host then.

PR #420, opened 2021-01-22
[host] nvfbc: update cursor position on shape change

quantum5:
This is because we keep track of the top-left corner of the cursor, not
the location of the hotspot. When the cursor shape changes, the hotspot
location may also change. When it does, the position of the top-left
corner changes and requires an update.

In the case that we do not have the current cursor position, which     
happens on startup, we do not generate the update.

### DIFF
--- a/host/platform/Windows/capture/NVFBC/src/nvfbc.c
+++ b/host/platform/Windows/capture/NVFBC/src/nvfbc.c
@@ -60,7 +60,7 @@ struct iface
   LGEvent * cursorEvents[2];
 
   int mouseX, mouseY, mouseHotX, mouseHotY;
-  bool mouseVisible;
+  bool mouseVisible, hasMousePosition;
 };
 
 static struct iface * this = NULL;
@@ -84,8 +84,9 @@ static void getDesktopSize(unsigned int * width, unsigned int * height, unsigned
 
 static void on_mouseMove(int x, int y)
 {
-  this->mouseX = x;
-  this->mouseY = y;
+  this->hasMousePosition = true;
+  this->mouseX           = x;
+  this->mouseY           = y;
   lgSignalEvent(this->cursorEvents[0]);
 }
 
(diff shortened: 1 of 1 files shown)


host/platform/Windows/capture/NVFBC/src/nvfbc.c
```diff
@@ -383,9 +384,10 @@ static int pointerThread(void * unused)
       this->mouseVisible = pointer.visible;
       this->mouseHotX    = pointer.hx;
       this->mouseHotY    = pointer.hy;
+      hotspotUpdated     = true;
     }
 
-    if (events[0])
+    if (events[0] || hotspotUpdated)
```
gnif, line 390: If a mouse move has never occurred since the host was started, the x/y position of the cursor will be invalid, this needs to be checked also.

gnif (changes requested):

host/platform/Windows/capture/NVFBC/src/nvfbc.c
quantum5: Done.

host/platform/Windows/capture/NVFBC/src/nvfbc.c
```diff
@@ -350,6 +350,8 @@ static CaptureResult nvfbc_getFrame(FrameBuffer * frame)
 
 static int pointerThread(void * unused)
 {
+  bool hasPosition = true;
```
gnif, line 353: Always true? This should be a member variable, and in `static void on_mouseMove(int x, int y)`

quantum5: Good catch, I meant false.

gnif: Even still, this should be set in `on_mouseMove`

quantum5: Done.

PR #432, opened 2021-01-28
[client] spice/wayland: improve cursor tracking logic

quantum5:
One of the major issues with the old tracking code is a data race
between the cursor thread updating g_cursor.guest and the
app_handleMouseBasic function. Specifically, the latter may have
sent mouse input via spice that has not been processed by the guest
and updated g_cursor.guest, but the guest may overwrite g_cursor.guest
to a previous state before the input is processed. This causes some
movements to be doubled. Eventually, the cursor positions will
synchronize, but this nevertheless causes a lot of jitter.

In this commit, we introduce a new field g_cursor.projected, which
is unambiguously the position of the cursor after taking into account
all the input already sent via spice. This is synced up to the guest
cursor upon entering the window and when the host restarts. Afterwards,
all mouse movements will be based on this position. This eliminates
all cursor jitter as far as I could tell.

Also, the cursor is now synced to the host position when exiting
capture mode.

A downside of this commit is that if the 1:1 movement patch is not
correctly applied, the cursor position would be wildly off instead
of simply jittering, but that is an unsupported configuration and
should not matter.

Also unsupported is when an application in guest moves the cursor
programmatically and bypassing spice. When using those applications,
capture mode must be on. Before this commit, we try to move the guest
cursor back to where it should be, but it's inherently fragile and
may lead to scenarios such as wild movements in first-person shooters.

### DIFF
--- a/client/src/main.h
+++ b/client/src/main.h
@@ -234,6 +234,9 @@ struct CursorState
 
   /* the guest's cursor position */
   struct CursorInfo guest;
+
+  /* the projected position after move, for app_handleMouseBasic only */
+  struct Point projected;
 };
 
 // forwards
--- a/client/displayservers/Wayland/wayland.c
+++ b/client/displayservers/Wayland/wayland.c
@@ -230,13 +230,11 @@ static const struct wl_registry_listener registryListener = {
 static void pointerMotionHandler(void * data, struct wl_pointer * pointer,
     uint32_t serial, wl_fixed_t sxW, wl_fixed_t syW)
 {
-  if (wm.relativePointer)
-    return;
-
   int sx = wl_fixed_to_int(sxW);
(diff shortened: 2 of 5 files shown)


client/src/main.h
```diff
@@ -234,6 +234,12 @@ struct CursorState
 
   /* the guest's cursor position */
   struct CursorInfo guest;
+
+  /* the projected position after move, for app_handleMouseBasic only */
+  struct {
+    int x;
+    int y;
+  } projected;
```
gnif, line 242: Please use `struct Point` as defined in `common/types.h`

gnif (changes requested):
Minor issue otherwise looks fine.

client/src/main.h
quantum5: Fixed.

PR #433, opened 2021-01-29
[host] windows: move log path to %ProgramData%\Looking Glass (host)

quantum5:
Instead of using %windir%\Temp, which is not accessible by default and
contains a lot of unrelated files, as the location for our log files,
this PR moves it to %ProgramData%\Looking Glass (host), which will
be a dedicated directory just for the LG host log files. This applies
to both the host application logs and the service logs.

Also, we now switched to using PathCombineA from shlwapi.dll instead
of using snprintf, which greatly simplifies the code. PathCombineA
guarantees that the path would not overflow a buffer of MAX_PATH.

Finally, this PR makes the installer create a shortcut to the log directory.

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -40,12 +40,9 @@ PASTE CLIENT OUTPUT HERE
 ```
 
 The entire (not truncated) log file from the host application (if applicable).
-To obtain this locate the log file on your system, it will be in one of the
-following two locations depending on how you are launching the Looking Glass Host
-application:
+Normally, this is found on the guest system at:
 
-  * C:\Windows\Temp\looking-glass.txt
-  * C:\Users\YOUR_USER\AppData\Local\Temp\looking-glass.txt
+    %ProgramData%\Looking Glass (host)\looking-glass-host.txt
 
 This log may be quite long, please delete the file first and then proceed to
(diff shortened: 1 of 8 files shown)


gnif:
Unfortunately this assumes the host will always be installed into the same directory, instead of hard coding to use the program files patth it would be better to get the executable's current directory and create a sub directory there for logging, ie:
```
- looking-glass-host.exe
- logs
  | - looking-glass-service.txt
  | - looking-glass-host.txt
```

gnif (changes requested):
As per my prior comment.

Edit: I misread the PR... it's going into Program Data, not Program Files... however when it's run as under the system user, where will this be?

quantum5:
There is only one ProgramData directory on the system, so if Windows is installed on `C:\`, this will be `C:\ProgramData\Looking Glass (host)`.

gnif:
Ah great! I thought this was per user. If that's the case one further change needs to be made, the help text in the client needs to be updated to reflect the new path.

https://github.com/gnif/LookingGlass/blob/c66cb493420a0563e40625093d07bc51a8c62000/client/src/main.c#L854

quantum5:
Done.

PR #437, opened 2021-01-30
[host] nvfbc: recreate nvfbc when recreation is requested

quantum5:
In the past, when NvFBCToSysCapture reports recreation is required, we
actually just do a reinitialization. This does not actually destroy
and recreate the NvFBC object. The result is an endless cycle of NvFBC
asking for recreation when it demands recreation. This commonly
manifests as the client waiting endlessly for the host when the guest
machine reboots.

In this commit, a CaptureResult value CAPTURE_RESULT_RECREATE is
introduced. When this result is received, the capture interface is
actually freed and recreated.


quantum5:
Superseded by #439.

PR #442, opened 2021-01-31
[client] egl: create 24-bit colour context

quantum5:
This should prevent the looking-glass-client window from having an alpha
channel. On Wayland, the alpha channel is used to compose the window onto
the desktop, so the wallpaper would bleed through unless set to complete
opaque.

We worked around this by using constant alpha for rendering, but it was
not sustainable. Instead, we should just ask for 24-bit context.

An alternative would to continue using 32-bit colour and the following
blend function:
glBlendFuncSeparate(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA, GL_ZERO, GL_ONE);

This approach is implemented in #443.

### DIFF
--- a/client/renderers/EGL/egl.c
+++ b/client/renderers/EGL/egl.c
@@ -530,7 +530,7 @@ bool egl_render_startup(void * opaque)
 
   EGLint attr[] =
   {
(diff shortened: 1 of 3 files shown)


gnif:
Works fine here

PR #443, opened 2021-01-31
[client] egl: ensure blended alpha is always opaque

quantum5:
When using glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA), the
resulting alpha channel is not fully opaque. On Wayland, the alpha
channel is used to compose the window onto the desktop, so the
wallpaper would bleed through if our window is not fully opaque.

glBlendFuncSeparate(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA, GL_ZERO, GL_ONE)
ensures that the alpha channel is opaque while blending colours as
before.

The alternative solution is #442.

### DIFF
--- a/client/renderers/EGL/alert.c
+++ b/client/renderers/EGL/alert.c
@@ -200,7 +200,7 @@ void egl_alert_render(EGL_Alert * alert, const float scaleX, const float scaleY)
     return;
 
   glEnable(GL_BLEND);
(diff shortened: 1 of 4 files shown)


gnif:
#442 was merged, is this still needed?

PR #444, opened 2021-01-31
[client] egl: fix race condition in help overlay

quantum5:
egl_help_set_text and egl_help_render were both accessing bmp->help from
different threads. This creates a race condition in which if the help text
is quickly toggled on and off, it stays on.

This has been fixed with an atomic exchange.

### DIFF
--- a/client/renderers/EGL/help.c
+++ b/client/renderers/EGL/help.c
@@ -24,6 +24,7 @@ Place, Suite 330, Boston, MA 02111-1307 USA
 #include "shader.h"
 #include "model.h"
 
+#include <stdatomic.h>
 #include <stdlib.h>
 #include <string.h>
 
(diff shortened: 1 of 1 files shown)


client/renderers/EGL/help.c
```diff
@@ -37,11 +38,12 @@ struct EGL_Help
...
-  EGL_Shader    * shader;
-  EGL_Shader    * shaderBG;
-  EGL_Model     * model;
-  LG_FontBitmap * bmp;
+  EGL_Texture * texture;
+  EGL_Shader  * shader;
+  EGL_Shader  * shaderBG;
+  EGL_Model   * model;
+
+  LG_FontBitmap * volatile _Atomic bmp;
```
gnif, line 46: `volatile` is not correct here, please use `_Atomic(LG_FontBitmap *) bmp;`
The atomic intrinsics handle cache coherency.

quantum5: Fixed.

PR #445, opened 2021-01-31
[client] spice: don't send zero deltas for Wayland input

Xyene:
While a compositor will never send us 0-delta motion events, they can
still end up as 0-deltas post-projection, consuming QEMU buffer space
for no reason.

This should help with mouse skipping issues.


Xyene:
This applies cleanly to master, might be worth backporting it there.

gnif:
Yes it would, can you please make this PR against master? I will rebase B3-next to include this fix to keep the history clean.

Xyene:
Done :+1:

PR #446, opened 2021-02-01
[client] spice: fix input:grabKeyboardOnFocus

quantum5:
It appears that at some point in the refactoring process, the logic for
grabKeyboardOnFocus was lost. This commit adds back the logic, allowing
grabKeyboardOnFocus to work once again.


gnif:
This was intentionally removed as it prevents one from grabbing the caption bar on Gnome/KDE to drag the window around. Keyboard grab should only happen when the cursor is in view & the window is focused.

PR #447, opened 2021-02-01
[client] spice: correctly ungrab keyboard when grabKeyboardOnFocus=no

quantum5:
When input:grabKeyboardOnFocus=no, exiting capture mode should ungrab
the keyboard. Otherwise, focusing the window doesn't grab the keyboard,
but toggling capture mode would leave the keyboard stuck in a grabbed
state until defocused.

### DIFF
--- a/client/src/main.c
+++ b/client/src/main.c
@@ -1563,7 +1563,7 @@ static void setGrabQuiet(bool enable)
   {
     if (params.grabKeyboard)
     {
-      if (!g_state.focused || params.captureInputOnly)
+      if (!params.grabKeyboardOnFocus || !g_state.focused || params.captureInputOnly)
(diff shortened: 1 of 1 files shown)


gnif:
> exiting capture mode

Exiting capture mode is not an event that causes loss of focus, I don't think this is the correct approach to this issue.

quantum5:
The point is that when `input:grabKeyboardOnFocus=no`, the keyboard should only be grabbed in capture mode.

But currently, when exiting capture mode, it does not ungrab the keyboard when `input:grabKeyboardOnFocus=no`, even though the keyboard was only grabbed when entering capture mode. So entering and exiting capture mode are not completely symmetric operations.

This PR does not alter the behaviour when `input:grabKeyboardOnFocus=yes`, which is the default.

PR #448, opened 2021-02-01
[client] renderer: add ability to toggle the FPS display

quantum5:
Before, if you want to see the FPS, you need to close the client and
restart it with the -k switch to see the FPS. This is annoying.

This PR introduces a new keybind, ScrollLock+P, which, when pressed,
toggles the display of the FPS.

This is implemented for both EGL and OpenGL backends.

### DIFF
--- a/client/src/keybind.c
+++ b/client/src/keybind.c
@@ -47,6 +47,12 @@ static void bind_video(int sc, void * opaque)
     core_startFrameThread();
 }
 
+static void bind_showFPS(int sc, void * opaque)
+{
+  g_state.showFPS = !g_state.showFPS;
+  app_showFPS(g_state.showFPS);
+}
+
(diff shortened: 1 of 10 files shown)


client/src/keybind.c
```diff
@@ -126,6 +132,7 @@ void keybind_register(void)
 {
   app_registerKeybind(KEY_F, bind_fullscreen, NULL, "Full screen toggle");
   app_registerKeybind(KEY_V, bind_video     , NULL, "Video stream toggle");
+  app_registerKeybind(KEY_P, bind_showFPS   , NULL, "FPS display toggle");
```
gnif, line 135: I think `KEY_D` would be a better option... `D` for Display.

quantum5: Changed to `KEY_D`. I have no strong feelings on the matter.

PR #450, opened 2021-02-01
README.md link maintenance

JJRcop:
* Added new Looking Glass Discord link (keeping VFIO link as well)
 * Level1Techs Triage thread replaced with Looking Glass board

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ https://looking-glass.io/downloads
 
 ## Web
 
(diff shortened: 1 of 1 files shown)


JJRcop:
Should `[doc]` be replaced with `[all]`?

gnif:
[doc] is correct for readme/documentation updates. There is no set standard for the tags in use, it's just to help whoever reads the commit log to quickly skip over commits they are not interested in.

PR #451, opened 2021-02-02
Fixing broken link

seffyroff:
Missing trailing slash from wiki link in Documentation section.


gnif:
NAK: The server configuration needs to be fixed

PR #453, opened 2021-02-03
Automatically disable screensaver when requested in the guest

quantum5:
In this PR, the host application is modified to detect if any application on the guest called `SetThreadExecutionState(ES_DISPLAY_REQUIRED | ES_CONTINUOUS)` to disable the screensaver and report this information to the client.

This PR also adds a new flag `win:autoScreensaver`, which when set to yes, automatically disables the screensaver on the host machine when we detect that the screensaver is disabled in the guest, and re-enables it when it is no longer disabled.

The new flag `win:autoScreensaver` conflicts with the flag `win:noScreensaver` (`-S`) and cannot be specified at the same time.

### DIFF
--- a/host/platform/Windows/src/platform.c
+++ b/host/platform/Windows/src/platform.c
@@ -25,6 +25,8 @@ Place, Suite 330, Boston, MA 02111-1307 USA
 #include <shellapi.h>
 #include <shlwapi.h>
 #include <fcntl.h>
+#include <powrprof.h>
+#include <ntstatus.h>
 
 #include "interface/platform.h"
 #include "common/debug.h"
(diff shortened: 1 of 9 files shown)


host/platform/Windows/src/platform.c
```diff
@@ -397,3 +399,24 @@ HWND os_getMessageWnd(void)
 {
   return app.messageWnd;
 }
+
+bool os_blockScreensaver()
+{
+  static bool      lastResult = false;
+  static ULONGLONG lastCheck  = 0;
+
+  ULONGLONG now = GetTickCount64();
```
gnif, line 408: Do we really need to poll this? Isn't there a windows power API we can register events for?

quantum5: Yes, there are power APIs that would happily give events for when things happen, like system suspend (`PowerRegisterSuspendResumeNotification`) and monitors turns on and off, power status, and whether the user activity timed out ([GUIDs](https://docs.microsoft.com/en-us/windows/win32/power/power-setting-guids) for PowerSettingRegisterNotification).

But I looked and looked, and I can't find *any* API other than `CallNtPowerInformation` that would report whether `SetThreadExecutionState` is used. I would gladly use an event-based API, but I can't find one.

gnif: Not a problem then, just making sure the option was explored first ;)

PR #455, opened 2021-02-05
[module] support in-memory kvmfr devices

quantum5:
Added an array option static_size_mb to the kvmfr module to create a
list of in-memory kvmfr devices. These devices support dmabuf just like
normal kvmfr devices. Additionally, they can be mmap'd, which allows
them to be passed to qemu as ivshmem devices.

This PR seems to work, but as someone with zero kernel experience, I don't necessarily trust it without careful review.

PR #456, opened 2021-02-06
[client] egl: rotate between multiple textures for DMA

quantum5:
This prevents glEGLImageTargetTexture2DOES from being called on a texture
while it's being rendered. Doing this would free the underlying data
structures for DMA during rendering, which may crash some drivers.

On radeonsi, this manifested as memory corruption. Valgrind reported that
memory in use by eglSwapBuffers was freed in glEGLImageTargetTexture2DOES.

This is solved by using multiple textures, one for each dmaFd. We reuse
the EGLImage for each dmaFd, preventing the destruction of images while
it is being used.

Along with #455, this appears to get dmabuf working on my VM→Host setup.

### DIFF
--- a/client/renderers/EGL/desktop.c
+++ b/client/renderers/EGL/desktop.c
@@ -45,6 +45,11 @@ struct DesktopShader
   GLint uCBMode;
 };
 
+struct DMAState {
+  int fd;
+  EGL_Texture * texture;
+};
+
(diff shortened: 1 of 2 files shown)


quantum5:
This has tearing issues. The implementation in #458 does not have such issues, so this PR has been superseded.

PR #458, opened 2021-02-07
[client] egl: force DMA copy into texture in on_frame

quantum5:
This commit forces the DMA'd memory to be copied into the texture in
the EGL on_frame handler. This avoids tearing when the LG host inevitably
updates the underlying memory. We need an additional copy inside the GPU,
but this is cheap compared to copying from system memory.

We could have used logic to lock the memory buffer, but that would require
performing DMA on every frame, which wastes memory bandwidth. This
manifests as reduced frame rate when moving the mouse compared to the
non-DMA implementation.

We also keep multiple EGLImages, one for each DMA fd, to avoid issues
with the OpenGL driver.

### DIFF
--- a/client/renderers/EGL/texture.c
+++ b/client/renderers/EGL/texture.c
@@ -79,6 +79,18 @@ struct EGL_Texture
   int             bufferCount;
   GLuint          tex;
   struct Buffer   buf[BUFFER_COUNT];
+
+  size_t dmaImageCount;
+  size_t dmaImageUsed;
+  struct
+  {
+    int fd;
+    EGLImage image;
+  }
+  * dmaImages;
+
+  GLuint dmaFBO;
+  GLuint dmaTex;
 };
 
 bool egl_texture_init(EGL_Texture ** texture, EGLDisplay * display)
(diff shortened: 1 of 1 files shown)


client/renderers/EGL/texture.c
```diff
@@ -79,6 +79,16 @@ struct EGL_Texture
   int             bufferCount;
   GLuint          tex;
   struct Buffer   buf[BUFFER_COUNT];
+
+  size_t dmaImageCount;
+  size_t dmaImageUsed;
+  struct {
```
gnif, line 85: nit: formatting
```c
struct
{
}
* dmaImages;
```

client/renderers/EGL/texture.c
```diff
@@ -365,41 +387,93 @@ bool egl_texture_update_from_dma(EGL_Texture * texture, const FrameBuffer * fram
...
+      {
+        DEBUG_EGL_ERROR("Failed to allocate memory");
+        eglDestroyImage(texture->display, image);
+        return false;
+      }
+      texture->dmaImageCount = newCount;
+      texture->dmaImages     = new;
+    }
+
+    size_t index = texture->dmaImageUsed++;
```
gnif, line 443: `const size_t`

gnif (changes requested):
Looks fine, just a few minor changes.

client/renderers/EGL/texture.c
quantum5: Fixed.

client/renderers/EGL/texture.c
quantum5: Fixed.

gnif (approved):

PR #457, opened 2021-02-06
[module] improvements after adding in-memory devices

quantum5:
This PR contains some improvements `mmap` implementations in the kvmfr module after #455.

This PR:

* implements mmaping for PCI `kvmfr` devices (not going through the `dmabuf`)
* uses more optimal implementation of `mmap` for `dmabuf`, bypassing the need to fault in the pages one by one
* improves the test program to catch offset handling issues in the `mmap` implementations
* provides sample correct output for the test program

PR #460, opened 2021-02-07
[common] ivshmem: do not create dmabuf for simple mmap

quantum5:
It used to be the case that you need to create dmabuf for kvmfr devices
to be able to mmap them. But after #457, this is no longer needed.

Directly mmaping the kvmfr device has the advantage of avoiding the
creation of a dmabuf, which has cost (e.g. the list of pages, the
scatterlist, etc.).

This PR should only be merged after #457.


gnif:
Looks good pending #457

gnif (approved):

PR #462, opened 2021-02-12
[client] wayland: implement asychronous clipboard read/writes

quantum5:
This PR implements asynchronous clipboard reads/writes.

To achieve this, it converts `waylandWait` to use `epoll(7)` and enables the registration and deregistration of additional file descriptors to wait for in the event loop. This machinery would be used for clipboard reads/writes.

With asynchronous operations came more opportunity for race conditions:
* A new clipboard offer may be received while we are handling the current one. We handle this by cancelling the current read and attempting to read the new clipboard offer.
* Multiple Wayland clients may request clipboard data, so we support multiple simultaneous outgoing transfers.
* Our clipboard data source may be cancelled while there are ongoing transfers. We continue those transfers until they are finished.

Since we need to continue clipboard transfers after the data source has been cancelled, we need to reference count the clipboard data. An reference counted buffer type has been introduced into the common library, in the `common/include/common/refcount.h` header. Once all references to the clipboard data has been removed, the data is freed.

This PR should be reviewed by @Xyene, who is the foremost expert on Wayland clipboard handling in the world.

### DIFF
--- a/client/displayservers/Wayland/wayland.c
+++ b/client/displayservers/Wayland/wayland.c
@@ -41,6 +41,7 @@ Place, Suite 330, Boston, MA 02111-1307 USA
 #include "app.h"
 #include "common/debug.h"
 #include "common/locking.h"
+#include "common/countedbuffer.h"
 
 #include "wayland-xdg-shell-client-protocol.h"
 #include "wayland-xdg-decoration-unstable-v1-client-protocol.h"
@@ -128,8 +129,7 @@ struct WaylandDSState
 
 struct WCBTransfer
 {
-  void * data;
-  size_t size;
+  struct CountedBuffer * data;
   const char ** mimetypes;
 };
 
@@ -1332,6 +1332,38 @@ static bool waylandCBInit(void)
   return true;
 }
 
+struct ClipboardWrite
+{
+  int fd;
+  size_t pos;
+  struct CountedBuffer * buffer;
+};
+
+static void clipboardWriteCallback(uint32_t events, void * opaque)
+{
+  struct ClipboardWrite * data = opaque;
+  if (events & EPOLLERR)
+    goto error;
+
+  ssize_t written = write(data->fd, data->buffer->data + data->pos, data->buffer->size - data->pos);
+  if (written < 0)
+  {
+    if (errno != EPIPE)
+      DEBUG_ERROR("Failed to write clipboard data: %s", strerror(errno));
+    goto error;
+  }
+
+  data->pos += written;
+  if (data->pos < data->buffer->size)
+    return;
+
+error:
+  waylandEpollUnregister(data->fd);
+  close(data->fd);
+  countedBufferRelease(&data->buffer);
+  free(data);
+}
+
 static void dataSourceHandleSend(void * data, struct wl_data_source * source,
     const char * mimetype, int fd)
 {
@@ -1343,19 +1375,19 @@ static void dataSourceHandleSend(void * data, struct wl_data_source * source,
     // wl-copy also stalls like this, but it's not necessary.
     fcntl(fd, F_SETFL, 0);
 
-    size_t pos = 0;
-    while (pos < transfer->size)
+    struct ClipboardWrite * data = malloc(sizeof(struct ClipboardWrite));
+    if (!data)
     {
-      ssize_t written = write(fd, transfer->data + pos, transfer->size - pos);
-      if (written < 0)
-      {
-        if (errno != EPIPE)
-          DEBUG_ERROR("Failed to write clipboard data: %s", strerror(errno));
-        goto error;
-      }
-
-      pos += written;
+      DEBUG_ERROR("Out of memory trying to allocate ClipboardWrite");
+      goto error;
     }
+
+    data->fd     = fd;
+    data->pos    = 0;
+    data->buffer = transfer->data;
+    countedBufferAddRef(transfer->data);
+    waylandEpollRegister(fd, clipboardWriteCallback, data, EPOLLOUT);
+    return;
   }
 
 error:
@@ -1366,7 +1398,7 @@ static void dataSourceHandleCancelled(void * data,
     struct wl_data_source * source)
 {
   struct WCBTransfer * transfer = (struct WCBTransfer *) data;
-  free(transfer->data);
+  countedBufferRelease(&transfer->data);
   free(transfer);
   wl_data_source_destroy(source);
 }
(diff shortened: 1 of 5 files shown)


client/displayservers/Wayland/wayland.c
```diff
@@ -47,6 +50,20 @@ Place, Suite 330, Boston, MA 02111-1307 USA
...
+
+typedef void (*WaylandPollCallback)(uint32_t events, void * opaque);
+
+struct WaylandPoll
+{
+  int fd;
+  bool removed;
+  WaylandPollCallback callback;
+  void * opaque;
+  struct WaylandPoll * prev;
```
Xyene, line 63: Use `struct wl_list link;` instead.

https://wayland.freedesktop.org/docs/html/apb.html

client/displayservers/Wayland/wayland.c
```diff
@@ -1031,28 +1198,90 @@ static const struct wl_data_offer_listener dataOfferListener = {
...
+  {
+    data->buf[data->numRead] = 0;
+    wcb.stashedType = data->type;
+    wcb.stashedSize = data->numRead;
+    wcb.stashedContents = data->buf;
+
+    waylandEpollUnregister(data->fd);
+    close(data->fd);
+    wl_data_offer_destroy(data->offer);
+    free(data);
```
Xyene, line 1243: Duped code with `clipboardReadCancel`?

client/displayservers/Wayland/wayland.c
quantum5: Switched to `wl_list`.

client/displayservers/Wayland/wayland.c
quantum5: This differs by the lack of `free(data->buf)`, but switched to `clipboardReadCancel` anyways, and added an extra flag to decide whether the buffer should be freed.

client/displayservers/Wayland/wayland.c
```diff
@@ -1062,43 +1277,35 @@ static void dataDeviceHandleSelection(void * data,
...
+  data->buf     = malloc(data->size);
+  data->offer   = offer;
+  data->type    = wcb.pendingType;
 
-      buf = nbuf;
-    }
+  if (!data->buf)
+  {
+    DEBUG_ERROR("Failed to allocate memory to receive clipboard data");
+    free(data);
```
Xyene, line 1297: Need to close the fd too.

client/displayservers/Wayland/wayland.c
```diff
@@ -1062,43 +1277,35 @@ static void dataDeviceHandleSelection(void * data,
...
   }
 
-  wcb.stashedSize = numRead;
-  wcb.stashedContents = buf;
+  if (!waylandEpollRegister(data->fd, clipboardReadCallback, data, EPOLLIN))
+  {
+    DEBUG_ERROR("Failed to register clipboard read into epoll: %s", strerror(errno));
+    free(data->buf);
+    free(data);
+  }
```
Xyene, line 1306: Same here, I think

client/displayservers/Wayland/wayland.c
```diff
@@ -1126,6 +1333,38 @@ static bool waylandCBInit(void)
...
+    if (errno != EPIPE)
+      DEBUG_ERROR("Failed to write clipboard data: %s", strerror(errno));
+    goto done;
+  }
+
+  data->pos += written;
+  if (data->pos < data->buffer->size)
+    return;
+
+done:
```
Xyene, line 1361: `error:` is probably the better name here.

Xyene (approved):
LGTM

common/CMakeLists.txt
```diff
@@ -19,6 +19,7 @@ set(COMMON_SOURCES
   src/option.c
   src/framebuffer.c
   src/KVMFR.c
+  src/refcount.c
```
gnif, line 22: Please rename this to `countedbuffer.c`

common/include/common/refcount.h
```diff
@@ -0,0 +1,30 @@
...
+
+#include <stddef.h>
+
+struct CountedBuffer {
+  _Atomic(size_t) refs;
+  size_t size;
+  char * data;
+};
+
+struct CountedBuffer * countedBufferCreate(size_t size);
```
gnif, line 28: Please use the following prototype for this instead, allowing the caller to provide the memory for `struct CountedBuffer` to avoid the need to double malloc. Failure should zero `return`.

    bool countedBufferNew(size_t size, struct CountedBuffer * return)

common/include/common/refcount.h
```diff
@@ -0,0 +1,30 @@
...
+void countedBufferIncRef(struct CountedBuffer * buffer);
+void countedBufferDecRef(struct CountedBuffer * buffer);
```
gnif, line 30: I am not sure I like the naming used here, I would suggest it follows that which is used by other referenced counting APIs like OLE/COM. Ie, `AddRef` and `Release`.

gnif (changes requested):

common/include/common/refcount.h
quantum5: Renamed to `countedBufferNew`.

Having the caller memory the memory for the `struct CountedBuffer` defeats the purpose of the entire thing, which is to allow the caller to call release and throw its reference away, while keeping the buffer around if something else wants it.

If you meant to have the caller `malloc` the structure, then have `countedBufferRelease` `free` the structure, I still think it's awkward, since the memory management is no longer handled in a single module.

common/CMakeLists.txt
quantum5: Renamed.

common/include/common/refcount.h
quantum5: Changed to COM-style `AddRef` and `Release`.

common/include/common/refcount.h
gnif: This is to allow a static buffer to be used/reused, the actual data member is still allocated via `malloc` by `counterBufferNew`, for example:

```
static struct CountedBuffer buffer;

int main()
{
  if (!countedBufferNew(1024, &buffer))
    return 1;
  return 0;
}
```

quantum5: This use case would not work. The intended use case is to keep past clipboard content around for ongoing transfers while the current reference is updated to the latest content. Statically allocating the `struct CountedBuffer` prevents us from such a use case, since we need multiple `struct CountedBuffer` with overlapping lifetimes.

gnif: Sorry yes, I follow its usage pattern now, I am just not keen on the double malloc. What about keeping the data contained within the struct?, ie:

```
struct CountedBuffer
{
struct CountedBuffer {
  _Atomic(size_t) refs;
  size_t size;
  // perhaps pad to ensure alignment, however this should be aligned anyway by gcc
  char data[];
};

data = malloc(sizeof(struct CountedBuffer) + size);
```

quantum5: Yea this sounds good. Changed to use the flexible array member.

common/src/countedbuffer.c
```diff
@@ -0,0 +1,44 @@
...
+  buffer->size = size;
+  return buffer;
+}
+
+void countedBufferAddRef(struct CountedBuffer * buffer)
+{
+  atomic_fetch_add(&buffer->refs, 1);
+}
+
+void countedBufferRelease(struct CountedBuffer * buffer)
```
gnif, line 40: `NULL` after free to make diagnostics easier with potential use after free bugs.

```
void countedBufferRelease(struct CountedBuffer ** buffer)
{
  if (atomic_fetch_sub(&(*buffer)->refs, 1) == 1)
  {
    free(*buffer);
    *buffer = NULL;
  }
}
```

quantum5: Done.

gnif (approved):

PR #463, opened 2021-02-13
[client] input: add releaseKeysOnFocusLoss option

Xyene:
This makes dealing with window manager shortcuts that overlap with guest
keys more pleasant, while retaining the previous functionality for users
who prefer it.

For instance, previously, using Alt+Tab (or $mod as Alt in i3/sway
movement commands) would result in the guest retaining Alt as pressed.

When the guest regained focus, it would continue thinking Alt is
pressed, leading to accidentally triggering obscure shortcuts. One had
to remember to press Alt again to "unstick" things, which was
suboptimal.

### DIFF
--- a/client/src/app.c
+++ b/client/src/app.c
@@ -57,6 +57,11 @@ void app_handleFocusEvent(bool focused)
   {
     core_setGrabQuiet(false);
     core_setCursorInView(false);
+
+    if (g_params.releaseKeysOnFocusLoss)
+      for (int key = 0; key < KEY_MAX; key++)
+        if (g_state.keyDown[key])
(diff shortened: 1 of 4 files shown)


client/src/app.c
```diff
@@ -57,6 +57,11 @@ void app_handleFocusEvent(bool focused)
   {
     core_setGrabQuiet(false);
     core_setCursorInView(false);
+
+    if (g_params.releaseKeysOnFocusLoss)
```
gnif, line 61: `if (!focused)` ?

gnif (approved):

PR #464, opened 2021-02-16
[client] ds: implement Wayland warp support

quantum5:
As it turns out, pointer confines can be used to warp cursors on the same surface on Wayland. This PR implements warps with this.

This may not necessarily work for all compositors. As such, the old cursor routines are still kept, and used when `wayland:warpSupport=no`.

This code is currently only tested to work on sway. Interested users should test this branch on Mutter and KWin and report findings.


Xyene:
This works on GNOME (Mutter). I was unable to try on Plasma 5.20; it's broken (LG fails to start due to missing interfaces) on 5.13, but 5.13 is ancient.

quantum5:
Tested on Plasma 5.20 and it works!

PR #465, opened 2021-02-18
[module] Add client ini example to kernel module readme

JJRcop:
An example to make using the module less painful


module/README.md
```diff
@@ -68,6 +68,13 @@ device node, for example:
 
     ./looking-glass-client -f /dev/kvmfr0
 
+Or, in `.looking-glass-client.ini`
```
gnif, line 71: `~/.looking-glass-client.ini` :)

PR #476, opened 2021-02-21
Add release type to CI matrix

luqasz:
Catch possible errors early in CI with release and debug builds.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,6 +6,9 @@ jobs:
     strategy:
       matrix:
         cc: [gcc, clang]
+        build_type:
+          - Release
+          - Debug
     steps:
     - uses: actions/checkout@v1
       with:
@@ -22,10 +25,12 @@ jobs:
           libx11-dev libxss-dev libxi-dev \
           wayland-protocols
     - name: Configure client
+      env:
+        CC: /usr/bin/${{ matrix.cc }}
       run: |
         mkdir client/build
         cd client/build
-        CC=/usr/bin/${{ matrix.cc }} cmake ..
+        cmake -DCMAKE_BUILD_TYPE={{ matrix.build_type }} ..
     - name: Build client
       run: |
         cd client/build


gnif:
Looks good, but please use the following commit message format:
`[section] part: description`
ie:
`[all] github: add release type to CI matrix`

.github/workflows/build.yml
```diff
@@ -25,7 +28,8 @@ jobs:
       run: |
         mkdir client/build
         cd client/build
-        CC=/usr/bin/${{ matrix.cc }} cmake ..
+        CC=/usr/bin/${{ matrix.cc }}
+        cmake -DCMAKE_RELEASE_TYPE={{ matrix.release_type }} ..
```
quantum5, line 32: It's actually `CMAKE_BUILD_TYPE`.

.github/workflows/build.yml
```diff
@@ -25,7 +28,8 @@ jobs:
       run: |
         mkdir client/build
         cd client/build
-        CC=/usr/bin/${{ matrix.cc }} cmake ..
+        CC=/usr/bin/${{ matrix.cc }}
```
quantum5, line 31: You can't put this on a separate line unless you prefix it with `export`. The CI build is currently using `/usr/bin/cc` as the C compiler for both GCC and Clang builds.

quantum5 (changes requested):

.github/workflows/build.yml
luqasz: Fixed.

.github/workflows/build.yml
luqasz: Fixed.

.github/workflows/build.yml
```diff
@@ -6,6 +6,9 @@ jobs:
     strategy:
       matrix:
         cc: [gcc, clang]
+        release_type:
```
quantum5, line 9: Maybe rename this? Otherwise looks fine.

luqasz: Isn't it already named `release_type` ? IMO self explanatory.

quantum5: It makes more sense to be called `build_type`.

gnif: Yes, `build_type` for consistency please.

luqasz: done

PR #478, opened 2021-02-22
[client] egl: make scale algorithms toggleable

quantum5:
The $escape+S keybinding now cycles through the available scale algorithms.
This allows the user to switch between algorithms if the automatic detection
turns out to be problematic.

The option is egl:scale. The scale algorithm number is displayed on the
alert to make it easier to specify the number for the option.


client/renderers/EGL/desktop.c
```diff
@@ -35,6 +35,9 @@ Place, Suite 330, Boston, MA 02111-1307 USA
 #include "desktop_rgb.frag.h"
 #include "desktop_rgh.def.h"
 
+#define LG_SCALE_AUTO    0
+#define SCALE_ALGORITHMS 3
```
gnif, line 39: Instead of splitting these out here, please define them in `desktop_rgb.def.h`, and if possible use an enum with a `MAX` element, ie:
```c
enum EGLScaleAlgo
{
  EGL_SCALE_AUTO,
  EGL_SCALE_NEAREST,
  EGL_SCALE_LINEAR,

  EGL_SCALE_MAX
};
```

Also as these defines are local only to EGL, so please use `EGL_` for them

client/renderers/EGL/desktop.c
```diff
@@ -35,6 +35,9 @@ Place, Suite 330, Boston, MA 02111-1307 USA
 #include "desktop_rgb.frag.h"
 #include "desktop_rgh.def.h"
```
gnif, line 36: I just noticed the header is misnamed, `rgb` vs `rgh`

client/renderers/EGL/desktop.c
```diff
@@ -61,6 +64,9 @@ struct EGL_Desktop
   // shader instances
   struct DesktopShader shader_generic;
 
+  // scale algorithm
+  int scaleAlgo;
```
gnif, line 68: `enum EGLScaleAlgo` or similar please.

gnif (changes requested):

client/renderers/EGL/desktop.c
quantum5: Okay, defined the `enum`. Unfortunately, it cannot be defined in that header since it's generated from the shader, and you can't define enums in GLSL. Therefore, we have to manually keep the shader constants in sync from now on.

client/renderers/EGL/desktop.c
quantum5: The file is deleted for reasons mentioned in the other comment, no action is necessary.

client/renderers/EGL/desktop.c
quantum5: Changed.

client/renderers/EGL/desktop.c
gnif: It would certainly be better to keep them defined in one place, which is why I stated, "if possible". So yes, keep defines but keep them all in the one place with a `_MAX` define.

quantum5: Done.

PR #482, opened 2021-02-24
Script that pulls some wiki pages into the repo

JJRcop:
Version 1 of a script utilizing `pandoc` to pull from the lg wiki and make MD files out of any page, then `sed` for some wiki link fixes (pandoc doesn't properly account for them)

Currently only configured for Installation and Frequently Asked Questions (and they both link to each other). If a page links to another wiki page without an md version configured, the md will link back to the wiki instead.

I didn't commit the results yet, but I did at https://github.com/JJRcop/LookingGlass/blob/pandoc-test/client/INSTALL.md. I will commit results to this PR if requested

Known issue: wiki page names replaced with md names don't have their tags updated to be md friendly

(Yes it does work, just being careful to not jump the gun if you want the files in different locations than they are first)

### DIFF
--- a/wiki-get
+++ b/wiki-get
@@ -22,19 +22,21 @@ WIKI_PREFIX_SED=$(
 	echo "$WIKI_PREFIX" | sed -e "s/\\//\\\\\//g" -e "s/\./\\\./g"
 )
 
-MD_WIKILINK_PAGE="[[:alpha:]_]+"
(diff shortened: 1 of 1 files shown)


JJRcop:
I hope that someone else can take a peek and give their thoughts, whether now or after it's merged. I openly admit to being a novice at this, had the sed manual open the whole time.

PR #487, opened 2021-02-25
[client] ds: remove SDL display server

quantum5:
The SDL display server is removed. Disabling both X11 and Wayland backends in CMake is made an error.

Also removed SDL-related language in the README and document the fact you can't disable both SDL and Wayland.

Finally, removed the comment about Wayland lacking warp support, as that has been rectified in #464.


gnif:
NAK: Removal of the DS, for now, is not wanted, this is still used to test behavioural differences between WMs

PR #504, opened 2021-03-02
[client] wayland: switch to using locked pointer when possible

quantum5:
Instead of using the confined pointer to do cursor capture, we instead
use locked pointers. We do not need the additional functionality that
confined pointer provides, and this also avoids compositor bugs that may
exist due to the more complicated implementation.

We still use confined pointers to do warp the pointer position.


quantum5:
The compositor bug in question has already been fixed, and this is no longer necessary.

PR #511, opened 2021-03-18
[host] windows: re-implement open log safely

quantum5:
Instead of doing ShellExecute from the service, we instead get the token
of the currently logged in user, and do CreateProcessAsUserA to run
notepad with that token. This should be safe.

### DIFF
--- a/host/platform/Windows/src/platform.c
+++ b/host/platform/Windows/src/platform.c
@@ -27,6 +27,8 @@ Place, Suite 330, Boston, MA 02111-1307 USA
 #include <fcntl.h>
 #include <powrprof.h>
 #include <ntstatus.h>
+#include <wtsapi32.h>
+#include <userenv.h>
 
 #include "interface/platform.h"
 #include "common/debug.h"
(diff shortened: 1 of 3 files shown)


host/platform/Windows/src/platform.c
```diff
@@ -86,6 +116,99 @@ static void RegisterTrayIcon(void)
...
+  char system32[MAX_PATH];
+  if (!GetSystemDirectoryA(system32, MAX_PATH))
+  {
+    DEBUG_WINERROR("Failed to get system directory", GetLastError());
+    return false;
+  }
+
+  if (!f_CreateProcessAsUserA && !windowsSetupAPI())
+  {
+    DEBUG_WINERROR("Failed to get CreateProcessAsUserA", GetLastError());
```
gnif, line 157: Instead of just erroring out, this should invoke the MsgBox showing the location. LG has Windows 8 users that this clearly would not work for.

quantum5: When this function errors out, it shows the message box, see line 249 below.

gnif: Ah yes, sorry... I missed that :)

gnif:
:+1:  This just needs rebasing

PR #512, opened 2021-03-18
[host] service: retry a few times when host fails to restart

quantum5:
Certain users of Radeon cards have observed that the host fails to start
at boot, with D3D11CreateDevice failing with HSTATUS 0x887a0004, which
translates to "The specified device interface or feature level is not
supported on this system."

This failure results in a LG_HOST_EXIT_FAILED exit code, which the service
does not attempt to restart. The user has to manually restart the service
for the host application to work.

These users reported that the host application started fine on
B2. This strongly suggests that the fix to enable capturing the login
screen made the host application start too early during the boot process,
and the graphics driver did not have time to initialize fully.

This PR allows the service to retry a few times on LG_HOST_EXIT_FAILED,
with exponential backoff, before giving up. This should cover this bug
and other similar bugs related to the early initialization which I do not
have logs for.

A preview version of the host is attached: [looking-glass-host-setup.zip](https://github.com/gnif/LookingGlass/files/6161931/looking-glass-host-setup.zip)

### DIFF
--- a/host/include/interface/platform.h
+++ b/host/include/interface/platform.h
@@ -29,6 +29,8 @@ Place, Suite 330, Boston, MA 02111-1307 USA
 #define LG_HOST_EXIT_KILLED  0x30
 // exit code for failed to start
 #define LG_HOST_EXIT_FAILED  0x40
+// exit code for failed to start, and no amount of restarting could help
+#define LG_HOST_EXIT_FATAL   0x50
 
 int  app_main(int argc, char * argv[]);
 bool app_init();
--- a/host/platform/Windows/src/service.c
+++ b/host/platform/Windows/src/service.c
@@ -41,6 +41,9 @@ Place, Suite 330, Boston, MA 02111-1307 USA
 #define SVC_ERROR ((DWORD)0xC0020001L)
 #define LOG_NAME  "looking-glass-host-service.txt"
 
+#define FAIL_MAX_RETRIES         5
+#define FAIL_RETRY_INIT_INTERVAL 1000
+
 /*
  * Windows 10 provides this API via kernel32.dll as well as advapi32.dll and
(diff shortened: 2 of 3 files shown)


quantum5:
Making this a draft for now until affected users could confirm if it fixes.

MasterNayru:
I have installed the host service attached to this PR on my machine and I am not seeing the connection issues on VM start that I was seeing before

quantum5:
Introduced fatal errors for things that no amount of restarting could fix.

For convenience, the new host executable is here: [looking-glass-host-setup.zip](https://github.com/gnif/LookingGlass/files/6167634/looking-glass-host-setup.zip)

quantum5:
> I have installed the host service attached to this PR on my machine and I am not seeing the connection issues on VM start that I was seeing before

Also can you send over `looking-glass-host-service.log`? I am curious as to how many retries were actually needed.

quantum5:
Also this prevents the host from exiting during driver updates.

MasterNayru:
@quantum5 Given that you have pushed a few commits since you last linked to a zip, should I wait for you to link to another one? I can't see in the pipeline output where I can get the produced build artifacts (if there is a download link in there somewhere, would prefer to get it from there myself)

quantum5:
@MasterNayru [looking-glass-host-setup.zip](https://github.com/gnif/LookingGlass/files/6167972/looking-glass-host-setup.zip)

MasterNayru:
Here are the logs after restarting a VM using that latest zip you attached: 
[Looking Glass logs (PR 512 #1).zip](https://github.com/gnif/LookingGlass/files/6168234/Looking.Glass.logs.PR.512.1.zip)

quantum5:
Seems like it restarted a maximum of 4 times for you, which suggests 5 attempts is probably the right number. Guess this PR is ready for review.

PR #516, opened 2021-03-28
Add command line option to constrain the client's window dimensions

ajruckman:
Add command line option to constrain the client's window dimensions to the guest's resolution when `dontUpscale` is also enabled.

I really like `dontUpscale`, but it is annoying that resizing the window above the guest's resolution makes black bars show up around the image.

When this option is enabled, the window is limited to the guest's resolution so you never get the black bars.


gnif (changes requested):
Looks good but please adjust your commit message to the following format and provide a summary as to what this change does. ie:

    [client] win: add new option win:shrinkOnUpscale

    This new option ... etc.

ajruckman:
Hi, thanks, I _think_ I've done this successfully...

PR #521, opened 2021-04-25
Do not leak fence on DMA mode

vroad:
Without `glDeleteSync`, memory usage of client slowly increases over time on DMA mode.


gnif:
Good find, but please update your commit message to the following format:
`[client] egl: do not leak fence when using DMABUF`

vroad:
I didn't see the message, I re-update the commit to update the message.

PR #528, opened 2021-05-01
[client] ui: add input:helpMenuDelay option

### LINKED ISSUE #527: RFE: New option `input:helpMenuDelay` for the escape key legend menu
Right now, the LG client waits for 0.2 seconds (or exactly 200000 µs) before showing the help menu when the escape key is held. This is too quick for my preferences, and I would prefer a time closer to 0.75 or even 1.

https://github.com/gnif/LookingGlass/blob/dec4c8533c252f29d90ad6598409305145f28c11/client/src/app.c#L400-L404

I propose a new option `input:helpMenuDelay`, a number in milliseconds (or, if you insist, microseconds) which configures the length of this delay.

quantum5:
This option controls the time period (in ms) after which the help menu
appears when holding down the escape key. After this time period,
capture mode is no longer toggled.

This fixes #527.

### DIFF
--- a/client/src/app.c
+++ b/client/src/app.c
@@ -397,7 +397,7 @@ void app_handleRenderEvent(const uint64_t timeUs)
   }
   else
   {
-    if (!g_state.escapeHelp && timeUs - g_state.escapeTime > 200000)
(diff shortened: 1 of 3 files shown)


JJRcop:
🤩 Thanks!

PR #530, opened 2021-05-03
[client] wayland: synchronize host cursor position with guest

quantum5:
This mirrors the x11 implementation, allowing the pointer to move
correctly into overlapping windows.

### DIFF
--- a/client/displayservers/Wayland/input.c
+++ b/client/displayservers/Wayland/input.c
@@ -413,3 +413,11 @@ void waylandRealignPointer(void)
   if (!wlWm.warpSupport)
     app_resyncMouseBasic();
 }
+
+void waylandGuestPointerUpdated(double x, double y, int localX, int localY)
+{
+  if (!wlWm.warpSupport || !wlWm.pointerInSurface)
+    return;
+
+  waylandWarpPointer(localX, localY, false);
+}
--- a/client/displayservers/Wayland/cursor.c
+++ b/client/displayservers/Wayland/cursor.c
@@ -106,7 +106,3 @@ void waylandShowPointer(bool show)
   wlWm.showPointer = show;
   wl_pointer_set_cursor(wlWm.pointer, wlWm.pointerEnterSerial, show ? wlWm.cursor : NULL, 0, 0);
 }
-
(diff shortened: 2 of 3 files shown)


client/displayservers/Wayland/input.c
```diff
@@ -413,3 +413,17 @@ void waylandRealignPointer(void)
   if (!wlWm.warpSupport)
     app_resyncMouseBasic();
 }
+
+void waylandGuestPointerUpdated(double x, double y, int localX, int localY)
+{
+  if (!wlWm.warpSupport || !wlWm.pointerInSurface || app_isCaptureMode())
```
quantum5, line 419: Not sure if the `app_isCaptureMode()` check should be done, as we currently can't block the cursor from entering overlapping windows in capture mode.

client/displayservers/Wayland/input.c
```diff
@@ -413,3 +413,17 @@ void waylandRealignPointer(void)
...
+    return;
+
+  static uint64_t last_warp = 0;
+  uint64_t now = microtime();
+  if (now - last_warp < 10000)
+    return;
+  last_warp = now;
```
quantum5, line 426: Not sure if this throttling is necessary. There appears to be no added CPU usage on sway from warping as far as I can tell from watching `htop`.

gnif: I needed to include this as rounding errors could cause multiple leave/enter events in X11 due to excessive warp commands in the X queue. It may not apply at all to Wayland.

client/displayservers/Wayland/input.c
gnif: Not sure for Wayland, X11 needs this to prevent window exit when in capture mode due to the trigger of the leave event when warping into a new window.

client/displayservers/Wayland/input.c
quantum5: It appears to be unnecessary. Removed.

client/displayservers/Wayland/input.c
quantum5: On Wayland, we are currently unable to prevent window exit in capture mode. A future solution using locked pointers should be able to fix this. I am removing the check so that the cursor at least exits predictably for now.

PR #537, opened 2021-05-12
[client] egl: use eglSwapBuffersWithDamageKHR for cursor-only updates

quantum5:
Instead of damaging the entire surface when rendering a cursor move,
we can use the EGL_KHR_swap_buffers_with_damage extension to only
damage the part of the window covered by the cursor. This should
reduce the cursor movement latency on Wayland.

This should probably be tested by some Wayland users first before merging.

### DIFF
--- a/client/displayservers/Wayland/gl.c
+++ b/client/displayservers/Wayland/gl.c
@@ -41,6 +41,7 @@ bool waylandEGLInit(int w, int h)
     DEBUG_ERROR("Failed to create EGL window");
     return false;
   }
+
   return true;
 }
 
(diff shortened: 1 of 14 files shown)


gasinvein:
This seems to work fine for me (on Fedora 34 with Gnome). Should I try to test something specific?

quantum5:
@gasinvein if there is a problem, it'll be really obvious: the cursor either fails to show up or leaves a broken trail behind.

client/displayservers/Wayland/gl.c
```diff
@@ -67,9 +68,61 @@ EGLDisplay waylandGetEGLDisplay(void)
...
+  {
+    const char *exts = eglQueryString(display, EGL_EXTENSIONS);
+    if (wl_proxy_get_version((struct wl_proxy *) wlWm.surface) < 4)
+    {
+      wlWm.eglSwapWithDamage = SWAP_WITH_DAMAGE_UNSUPPORTED;
+      DEBUG_INFO("Swapping buffers with damage: not supported, need wl_compositor v4");
+    }
+    else if (strstr(exts, "EGL_KHR_swap_buffers_with_damage") && g_egl_dynProcs.eglSwapBuffersWithDamageKHR)
+    {
+      wlWm.eglSwapWithDamage = SWAP_WITH_DAMAGE_KHR;
```
gnif, line 83: The prototypes of KHR and EXT look identical (unless I missed something), why not store the function pointer at walyandInit to avoid these conditionals running on every call?

quantum5: We are unable to query the EGL extensions without a context, which is created in the `EGL` renderer. The only place we are guaranteed to have an EGL context active is during the `waylandEGLSwapBuffers` call.

quantum5:
It's probably best to change this PR to use `util_hasGLExt` after #538.

client/displayservers/Wayland/gl.c
gnif: Ok, but even still there is an additional check on the enum below to decide which method to call, why not store the pointer to the correct one and just call it?

quantum5: Good idea, changed.

gnif:
> It's probably best to change this PR to use `util_hasGLExt` after #538.

Agreed

quantum5:
This should be ready for review now.

PR #538, opened 2021-05-13
[client] egl: use new util_hasGLExt helper to check extensions

quantum5:
We previously used strstr, which can be prone to false positives when
the name of one extension is a substring of another extension.

This commit creates the helper function util_hasGLExt, which asserts
that the substring found in extension list is bounded by either spaces
or the beginning/end of the string.

### DIFF
--- a/client/src/util.c
+++ b/client/src/util.c
@@ -21,9 +21,11 @@ Place, Suite 330, Boston, MA 02111-1307 USA
 #include "main.h"
 
 #include "common/debug.h"
+#include "common/stringutils.h"
 
 #include <stdlib.h>
 #include <stdio.h>
+#include <string.h>
 #include <assert.h>
 #include <math.h>
 
@@ -206,4 +208,9 @@ void util_rotatePoint(struct DoublePoint *point)
       point->y =  temp;
       break;
   }
+}
+
+bool util_hasGLExt(const char * exts, const char * ext)
+{
+  return str_containsValue(exts, ' ', ext);
 }
--- a/common/include/common/stringutils.h
+++ b/common/include/common/stringutils.h
@@ -20,6 +20,8 @@ Place, Suite 330, Boston, MA 02111-1307 USA
 #ifndef _H_LG_COMMON_STRINGUTILS
 #define _H_LG_COMMON_STRINGUTILS
 
+#include <stdbool.h>
+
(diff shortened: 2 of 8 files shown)


client/src/util.c
```diff
@@ -207,3 +208,22 @@ void util_rotatePoint(struct DoublePoint *point)
       break;
   }
 }
+
+bool util_hasGLExt(const char * exts, const char * ext)
```
gnif, line 212: This is simply a word-based text search, perhaps `util_containsWord` or a similar non-GL specific name?

quantum5: This function is not really a general word search function because it only works on space separated lists. If it was called `util_containsWord`, I would expect it to handle punctuation, which is not the case here.

It does not make sense to generalize this function because OpenGL extension lists are separated only by spaces, so I think this function should stay as is.

gnif: true, however, I am not keen on adding GL-specific methods (even if just in name) to the `common` project, it's intended to be generic utility functions and code that is not specific to any particular technology. IMO it would be better to name this `util_containsValue` with the delimiter being an argument making this a generic string method, ie:

`bool util_containsValue(const char * list, const char delimiter, const char * value)`

quantum5: This is in the client project, not the common project.

gnif: Ah, you're correct, in which case IMO it should be in `common/src/stringutils.c` and then have the GL specific version in utils.c which calls the generic version in stringutils.

gnif: btw, the reason being that the OBS client or even the host may at some point find use in this method, best to avoid code duplication.

common/include/common/stringutils.h
```diff
@@ -28,4 +30,7 @@ int valloc_sprintf(char ** str, const char * format, va_list ap)
 int alloc_sprintf(char ** str, const char * format, ...)
   __attribute__ ((format (printf, 2, 3)));
 
+// Find value in a list separated by delimiter.
+bool util_containsValue(const char * list, char delimiter, const char * value);
```
gnif, line 34: Sorry, but one last thing, this is no longer part of the `util_` function set, perhaps `str_`?

quantum5: Done.

PR #542, opened 2021-05-22
[obs] fix oob cursor writes

arcnmx:
Allocate enough memory for the cursor data!

I think the alpha mask is still messed up for monochrome cursors but at least it doesn't segfault now.

### DIFF
--- a/obs/lg.c
+++ b/obs/lg.c
@@ -259,7 +259,7 @@ static void * pointerThread(void * data)
 
         case CURSOR_TYPE_MONOCHROME:
         {
-          dataSize = cursor->height * sizeof(uint32_t);
+          dataSize = cursor->height * cursor->width * sizeof(uint32_t);
           allocCursorData(this, dataSize);
 
           const int hheight = cursor->height / 2;


gnif:
I will have look closer at this however the fix proposed is certainly incorrect, if anything it should be `width * 2 * sizeof(uint32_t)` as windows monochrome cursors are bitmasks using the bits as the columns. See: https://docs.microsoft.com/en-us/windows/win32/menurc/using-cursors

arcnmx:
I should dig into it a bit more but... AFAICT the memory being allocated is for the resulting GL texture (assigned to `d` in the loop before), not the original data. Which as far as I can tell is `hheight * width * 2` pixels in size, x32bits?

gnif:
Upon further review you're correct, however, the fix is still not quite right. It should be `dataSize = cursor->height * cursor->width * cursor->pitch;` for correctness.

arcnmx:
Hm, confusing. `cursor->pitch` is only used to index into the src data, and doesn't seem to influence the new converted texture layout (since `pitch` is presumably just `width/8` rounded up or something)? Our newly allocated `dest_pitch` is presumably still equivalent to `cursor->width * sizeof(u32)` - assuming there are power-of-2 restrictions on these textures or other layout requirements I'm not yet aware of.
In other words if `pitch` is correct to use for the allocation size, it should also be used when indexing into `d` to initialize the data, which it currently isn't used for.

gnif:
Again, yes you're correct, I am letting my dyslexia get the better of me.

Re blending, I am not great with OpenGL but I do know that if you have layering in OBS it often helps to put LG on it's own Scene and then include that scene.

arcnmx:
The blending thing is me seeing this: 2021-05-22-192126_49x48_scrot but I think this one might be a masked color cursor instead. I was seeing this before patching anything, so is an unrelated issue presumably.

The I-beam cursor on the other hand no longer causes a segfault, and seems to render its xor blending just fine.

PR #552, opened 2021-05-30
[client] wayland: offer all supported clipboard formats

quantum5:
This commit restructures the Wayland clipboard handling for host->VM.

Before, we select one clipboard format and buffers the data for it, to
be presented to spice when needed.

Now, we simply offer all clipboard formats supported, and only when spice
asks for the data do we actually read the wl_data_offer. The wl_data_offer
is kept around until a new offer is presented, the offer invalidated, or
when we lose keyboard focus. This is in accordance with the specification
for wl_data_device::selection, which states that:

> The data_offer is valid until a new data_offer or NULL is received or
> until the client loses keyboard focus. The client must destroy the
> previous selection data_offer, if any, upon receiving this event.

We still buffer the entire clipboard data into memory because we have no
knowledge of the clipboard data size in advance and cannot do incremental
transfers.

Furthermore, if the user performs drag-and-drop on our window, we may have
need to handle multiple wl_data_offer objects at the same time. Therefore,
instead of storing state on the global wlCb object, we instead allocate
memory and store it as user_data on the wl_data_offer. As a result, we also
handle drag-and-drop so that we can free the memory.

Currently, this appears to work, but it's best that @Xyene review this.

### DIFF
--- a/client/displayservers/Wayland/clipboard.c
+++ b/client/displayservers/Wayland/clipboard.c
@@ -33,6 +33,11 @@ Place, Suite 330, Boston, MA 02111-1307 USA
 #include "app.h"
 #include "common/debug.h"
 
+struct DataOffer {
+  bool isSelfCopy;
+  char * mimetypes[LG_CLIPBOARD_DATA_NONE];
+};
+
 static const char * textMimetypes[] =
 {
   "text/plain",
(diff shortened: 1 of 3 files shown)


gnif:
Please let me know when it's been checked by @Xyene that it's OK.

Xyene (approved):
Reviewed out-of-band, LGTM now.

PR #553, opened 2021-05-31
[obs] install into plugin dir

arcnmx:
`/usr/share/obs/blahblah` is where obs expects to find the dylib, so put it there!

### DIFF
--- a/obs/CMakeLists.txt
+++ b/obs/CMakeLists.txt
@@ -32,6 +32,23 @@ add_compile_options(
 set(EXE_FLAGS "-Wl,--gc-sections")
 set(CMAKE_C_STANDARD 11)
 
+if(CMAKE_SIZEOF_VOID_P EQUAL 8)
+	set(OBS_PLUGIN_DIR 64bit)
+else()
+	set(OBS_PLUGIN_DIR 32bit)
+endif()
+
+if(USER_INSTALL)
+	if(DEFINED ENV{XDG_CONFIG_HOME})
+		set(OBS_PLUGIN_PREFIX $ENV{XDG_CONFIG_HOME})
+	else()
+		set(OBS_PLUGIN_PREFIX $ENV{HOME}/.config)
+	endif()
+	set(OBS_PLUGIN_PREFIX ${OBS_PLUGIN_PREFIX}/obs-studio/plugins)
+else()
+	set(OBS_PLUGIN_PREFIX ${CMAKE_INSTALL_DATADIR}/obs/obs-plugins)
+endif()
+
 add_definitions(-D ATOMIC_LOCKING)
 get_filename_component(PROJECT_TOP "${PROJECT_SOURCE_DIR}/.." ABSOLUTE)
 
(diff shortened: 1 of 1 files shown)


obs/CMakeLists.txt
```diff
@@ -18,20 +18,27 @@ endif()
 option(ENABLE_BACKTRACE "Enable backtrace support on crash" ON)
 add_feature_info(ENABLE_BACKTRACE ENABLE_BACKTRACE "Backtrace support.")
 
-add_compile_options(
+set(CMAKE_POSITION_INDEPENDENT_CODE ON)
```
gnif, line 21: Please split this up into separate commits, this has nothing to do with the installation path.

obs/CMakeLists.txt
```diff
@@ -18,20 +18,27 @@ endif()
...
   "$<$<CONFIG:DEBUG>:-O0;-g3;-ggdb>"
 )
 
 set(EXE_FLAGS "-Wl,--gc-sections")
 set(CMAKE_C_STANDARD 11)
 
+if(CMAKE_SIZEOF_VOID_P EQUAL 8)
+	set(OBS_PLUGIN_DIR 64bit)
+else()
+	set(OBS_PLUGIN_DIR 32bit)
```
gnif, line 39: Do we even care about 32-bit, LG would be practically useless on such a system.
Also separate commit please

obs/CMakeLists.txt
```diff
@@ -63,10 +59,15 @@ add_subdirectory("${PROJECT_TOP}/common"          "${CMAKE_BINARY_DIR}/common")
...
 add_library(looking-glass-obs SHARED ${SOURCES})
-target_link_libraries(looking-glass-obs
+target_compile_options(looking-glass-obs PRIVATE ${C_FLAGS})
+target_link_libraries(looking-glass-obs PRIVATE
 	${EXE_FLAGS}
 	lg_common
 	lgmp
 )
 
+install(TARGETS looking-glass-obs
```
gnif, line 69: what about a local install into the local obs plugins folder for the user?

gnif (changes requested):

obs/CMakeLists.txt
arcnmx: Unless I messed up these should be separate commits, though I can pull it out into another PR if you'd prefer.

obs/CMakeLists.txt
arcnmx: Dunno, but seems better than hard-coding 64bit. Separate commit from what, the `install()` or the cflags?

obs/CMakeLists.txt
arcnmx: Good point, that might make sense as an option... `${CMAKE_INSTALL_DATADIR}/obs/obs-plugins/` becomes `$ENV{HOME}/.config/obs-studio/plugins/` I think?

obs/CMakeLists.txt
gnif: No, this is part of the one commit, please split it up.

obs/CMakeLists.txt
gnif: part of the install changes sorry, and yeah, keep this, it's fine :)

obs/CMakeLists.txt
arcnmx: I just removed the second commit from this PR, I'll continue to experiment with my local install in [another branch](https://github.com/arcnmx/LookingGlass/compare/obs-cmake...arcnmx:obs-cmake2). None of it is all that relevant anyway, I was just trying to clean up a bit and tie the flags to targets to make it easier to customize.

PR #564, opened 2021-06-06
[client] wayland: lock confine-related code to avoid race

quantum5:
This should fix the occasional Wayland protocol errors that arise when
the UI thread and the cursor thread race.

Example of error that is fixed:

    zwp_pointer_constraints_v1@11: error 1: a pointer constraint with a wl_pointer of the same wl_seat is already on this surface


quantum5:
I haven't seen the crash for two days, so I think this fixed the race.

PR #577, opened 2021-07-02
[host] avoid manual windres command

arcnmx:
This causes cmake to automatically pick up the right program via `CMAKE_RC_COMPILER` from the toolchain config and use the correct RCFLAGS and so on.


gnif:
Was this tested on both Windows and Linux? IIRC I built it this way because it had issues finding the executable under windows.

arcnmx:
The native CI result seems to be okay, if it doesn't work that sounds like an improperly configured build environment? cmake is supposed to already know what executable to use via CMAKE_RC_COMPILER, provided by the toolchain config or via RC env var etc.

PR #579, opened 2021-07-05
[host] avoid manual windres command

arcnmx:
Not entirely sure what's up with the revert, but I'm assuming the issue was that by including the .rc.o in the platform .a, the linker wasn't picking it up? I think you'd need to use `--whole-archive` flag (or `OBJECT` library targets or something) to resolve it in that way, but a less intrusive change seems like just adding it directly as an object on the executable as I do so here.

### DIFF
--- a/host/CMakeLists.txt
+++ b/host/CMakeLists.txt
@@ -66,7 +66,10 @@ add_subdirectory("${PROJECT_TOP}/repos/LGMP/lgmp" "${CMAKE_BINARY_DIR}/lgmp"  )
 add_subdirectory(platform)
 
 if(WIN32)
-  add_executable(looking-glass-host WIN32 ${SOURCES})
+  add_executable(looking-glass-host WIN32
+    platform/Windows/resource.rc
+    ${SOURCES}
+  )
 else()
   add_executable(looking-glass-host ${SOURCES})
 endif()
--- a/host/platform/Windows/CMakeLists.txt
+++ b/host/platform/Windows/CMakeLists.txt
@@ -11,6 +11,7 @@ add_library(platform_Windows STATIC
 	src/mousehook.c
 	src/force_compose.c
 	src/delay.c
+	resource.rc
 )
 
 # allow use of functions for Windows Vista or later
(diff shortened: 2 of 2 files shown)


gnif:
No, the object was in use, the problem is `winders is running with its CWD as the output directory, causing it to fail to find the `icon.ico` and `app.manifest` files.

arcnmx:
> No, the object was in use, the problem is windres is running with its CWD as the output directory, causing it to fail to find the icon.ico and app.manifest files.

Hm, I've noticed no issues with it finding the resources locally nor with the github actions CI. Do you have build logs of this failing, ideally cmake's verbose build logs?

What I did see is that the linker may not actually pull the compiled object in, and thus you'd get a successful compile with a missing icon/manifest, but that was because of the static linking --whole-archive stuff that 64170af works around.

(it always worked previously because the object was listed as a `target_link_library`, which happened to result in the linker not trying to discard it)

> As the build directory can be literally anywhere this is never going to work unless cmake can be made to run windres with the CWD as the source directory.

This doesn't make sense to me, afaik cmake knows to provide include directories to help it find resources like it would when compiling any other file out-of-source-tree. It also does cd before building the file iirc, which I think you can specify via source file property, but I'd need to double check that. cmake does know how to run windres anyway, or at least it's supposed to!

(In fact I might recommend adding `${PROJECT_TOP}/resources` to the include dirs so that the rc file can just specify "icon.ico" alone without using the relative path, but it works with it so whatever)

quantum5:
Created #620 which should address the issues in this PR.

arcnmx:
I'm not sure what those issues are or if there even are any..?

quantum5:
Actually your PR works as is for building `resources.rc`. There is an issue of it not rebuilding when `app.manifest` is updated, but that was a problem in the custom command version as well.

arcnmx:
Mm the use of `OBJECT_DEPENDS` is a good idea and fixes that quite well! I also like the removal of the `../../` include though I believe that's just a cosmetic change because it should still be working regardless.

quantum5:
I am making #620 a change based on this PR and will only be covering the `OBJECT_DEPENDS` part for simplicity, as otherwise we would have two PRs that do sort of similar things.

PR #582, opened 2021-07-06
[doc] index: Nuke trello and add ethereum address

JJRcop:
The LG Trello is unmaintained, and we have to fund those rare NFTs somehow.


JJRcop:
Corrected log file tray operation, since LG opens the log file securely rather than providing the location.

PR #583, opened 2021-07-09
[doc] module: dkms example has explicit "."

JJRcop:
This works in dash (not just bash), so should be portable to most shells.
Will hopefully help users not misinterpret the period as a mistake.


JJRcop:
The bash highlighting also makes it show up green, which might help too 
Screenshot_20210708_232523

PR #598, opened 2021-07-18
[client] imgui: support registering new graphs

quantum5:
Currently, our struct ll doesn't support removing elements from the middle,
so we will not be removing anything for now.


quantum5:
Superseded by #599.

PR #620, opened 2021-07-20
[host] windows: fix resource compilation dependencies

quantum5:
It used to be the case that when updating app.manifest, the resource file
it not automatically rebuilt. This made it a headache to update the manifest.
We set OBJECT_DEPENDS so that cmake knows to make the res file depend on
app.manifest and icon.ico.

This commit is based on PR #579 and should be rebased on it after it's merged.

### DIFF
--- a/host/CMakeLists.txt
+++ b/host/CMakeLists.txt
@@ -71,6 +71,9 @@ add_subdirectory("${PROJECT_TOP}/repos/LGMP/lgmp" "${CMAKE_BINARY_DIR}/lgmp"  )
 add_subdirectory(platform)
 
 if(WIN32)
+  set_source_files_properties(platform/Windows/resource.rc PROPERTIES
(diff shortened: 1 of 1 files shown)


host/CMakeLists.txt
```diff
@@ -66,7 +66,14 @@ add_subdirectory("${PROJECT_TOP}/repos/LGMP/lgmp" "${CMAKE_BINARY_DIR}/lgmp"  )
 add_subdirectory(platform)
 
 if(WIN32)
-  add_executable(looking-glass-host WIN32 ${SOURCES})
+  set(CMAKE_RC_FLAGS "-I${PROJECT_TOP}")
```
arcnmx, line 69: IIRC you can just use a normal `target_include_directories` to do this in a more portable way and not clobber the `CMAKE_RC_FLAGS` var. source properties can work too.

quantum5: For simplicity's sake, I am making this PR just do the `OBJECT_DEPENDS` stuff.

PR #622, opened 2021-07-20
[host] windows: directly invoke ChangeWindowMessageFilterEx

quantum5:
This requires bumping the minimum Windows version to Windows 7, but I
don't think we care about Vista anyways.


gnif:
We have a few windows 7 users still that are using NvFBC, this was done this way specifically to remain compatible.

quantum5:
Windows 7 is still supported after this PR. Windows Vista will not, but is anyone still using it?

gnif:
Oh, lol, I read `Windows 8`, lol.

PR #661, opened 2021-07-26
fps limiter option

arcnmx:
In scenarios where the guest monitor is used as a primary display (for example when combined with the obs client), the unnecessary strain on the GPU/CPU to capture all frames unbounded is excessive.

### DIFF
--- a/host/src/app.c
+++ b/host/src/app.c
@@ -127,6 +127,13 @@ static struct Option options[] =
     .value.x_string = "",
     .validator      = validateCaptureBackend,
   },
+  {
+    .module         = "app",
+    .name           = "throttleFPS",
+    .description    = "Throttle Capture Frame Rate",
+    .type           = OPTION_TYPE_INT,
+    .value.x_int    = 0,
+  },
   {0}
 };
 
@@ -608,6 +615,10 @@ int app_main(int argc, char * argv[])
     }
   }
 
+  int throttleFps = option_get_int("app", "throttleFPS");
+  int throttleUs = throttleFps ? 1000000 / throttleFps : 0;
+  uint64_t previousFrameTime = 0;
+
   const char * ifaceName = option_get_string("app", "capture");
   CaptureInterface * iface = NULL;
   for(int i = 0; CaptureInterfaces[i]; ++i)
@@ -720,6 +731,16 @@ int app_main(int argc, char * argv[])
         sendPointer(true);
         LG_UNLOCK(app.pointerLock);
       }
+
+      const uint64_t now = microtime();
+      const uint64_t delta = now - previousFrameTime;
+      if (delta < throttleUs)
+      {
+        nsleep((throttleUs - delta) * 1000);
+        previousFrameTime = microtime();
+      }
+      else
+        previousFrameTime = now;
 
       switch(iface->capture())
       {


gnif:
As much as this might sound like a good idea, system performance suffers due to this under many titles due to the way Microsoft has implemented DXGI DD. Please see the information regarding accumulated updates for `DXGIOutputDuplication::ReleaseFrame`. While we do hold the frame and release it just before the next capture as Microsoft recommends, the behaviour on differing systems is inconsistent.

quantum5:
It doesn't make sense to use looking glass to record with OBS while looking at the guest monitor. You will almost certainly get better results running OBS inside the VM. You will be able to avoid the IVSHMEM overhead. OBS plugin only makes sense when you are using the looking glass client and paying for the capture already.

gnif:
Further to this please note that `sleep` and its family under Windows has very poor accuracy, this is not a stable method to limit the framerate. You may have better results with the util function we added recently for windows `delayExecution` which uses `NtDelayExecution` for better accuracy.

https://github.com/gnif/LookingGlass/blob/master/host/platform/Windows/include/windows/delay.h

Further reading: https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-2000-server/cc938613(v=technet.10)?redirectedfrom=MSDN

arcnmx:
> Further to this please note that sleep and its family under Windows has very poor accuracy, this is not a stable method to limit the framerate. You may have better results with the util function we added recently for windows delayExecution which uses NtDelayExecution for better accuracy.

Yeah I didn't want to complicate the change with platform-specific considerations but I did notice that and initially wanted to use `nanosleep`, but it wasn't implemented. Would it maybe make sense to move `delayExecution` into common/time instead so it can be called by the common interface?

gnif:
There is nothing here to reject other than a few code formatting issues and the sleep issue. I will see about implementing `nanosleep` for Windows that wraps `NttDelayExecution` shortly. In the meantime, I will review the code as-is and provide feedback.

host/src/app.c
```diff
@@ -721,6 +731,15 @@ int app_main(int argc, char * argv[])
         LG_UNLOCK(app.pointerLock);
       }
 
+      const uint64_t now = microtime();
+      const uint64_t delta = now - previousFrameTime;
+      if (delta < throttleUs) {
```
gnif, line 736: `{` on a seperate line please.

host/src/app.c
```diff
@@ -721,6 +731,15 @@ int app_main(int argc, char * argv[])
...
+        usleep(throttleUs - delta);
+        previousFrameTime = microtime();
+      } else {
```
gnif, line 739: ditto here, format for this project is:
```c
}
else
{
```

for single line if statements, no braces are needed unless the code is hard to follow without them.

host/src/app.c
```diff
@@ -608,6 +615,9 @@ int app_main(int argc, char * argv[])
     }
   }
 
+  int throttleFps = option_get_int("app", "throttleFPS");
+  int throttleUs = throttleFps ? 1000000 / throttleFps : 0;
+  uint64_t previousFrameTime = 0;
   const char * ifaceName = option_get_string("app", "capture");
```
gnif, line 621: Line break after this please.

host/src/app.c
```diff
@@ -127,6 +127,13 @@ static struct Option options[] =
     .value.x_string = "",
     .validator      = validateCaptureBackend,
   },
+  {
+    .module         = "app",
+    .name           = "throttleFPS",
+    .description    = "Throttle capture FPS",
```
gnif, line 133: `Throttle Capture Frame Rate` please.

gnif (commented):
Will let you know when nanosleep is ready for use.

gnif:
done, please use `nsleep` as defined in `common/time.h` which wraps `NtDelayExecution`

host/src/app.c
```diff
@@ -721,6 +732,18 @@ int app_main(int argc, char * argv[])
...
 
+      const uint64_t now = microtime();
+      const uint64_t delta = now - previousFrameTime;
+      if (delta < throttleUs)
+      {
+        nsleep((throttleUs - delta) * 1000);
+        previousFrameTime = microtime();
+      }
+      else
+      {
```
gnif, line 743: single-line if/else statement, please use the format
```
if (x)
{
  statement;
  statement;
 }
else
  statement;
```

PR #662, opened 2021-07-26
[host] nvfbc: fix frame size for rotated displays

### LINKED ISSUE #555: Rotations confuse and crash NvFBC capture
Mixing NvFBC with a portrait rotation causes an endless crash loop in the host application, and no frames ever get through. I'm currently investigating / digging into it / trying to get logs, but am assuming it's trivially reproducible and that it's some mismatch between the capture vs monitor width/height. DXGI on the other hand seems to handle the rotation fine.

The host logs also claim that the capture frame size (and required ivshmem size) is significantly higher than what's normally reported when in landscape mode, possibly a sign that the frame stride size is wrong?

arcnmx:
Instead we can just get the frame size directly from NvFBC rather than WinAPI.

Fixes #555

### DIFF
--- a/host/platform/Windows/capture/NVFBC/src/nvfbc.c
+++ b/host/platform/Windows/capture/NVFBC/src/nvfbc.c
@@ -84,19 +84,6 @@ static bool nvfbc_deinit(void);
 static void nvfbc_free(void);
 static int pointerThread(void * unused);
 
-static void getDesktopSize(unsigned int * width, unsigned int * height)
-{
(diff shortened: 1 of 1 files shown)


gnif:
Sorry, but NvFBC lies about the frame size when the resolution is changed but the driver decides not to re-init the frame buffer. This code is here for a reason.

arcnmx:
Hm, did you ever ingestigate whether throwing a `CAPTURE_RESULT_REINIT` when the resolution changes works instead or does that not even reinit the framebuffer properly?

gnif:
Yeah, we don't even know the res changed... it's likely this is why the members are `maxWidth` and `maxHeight, implying it's the GPU frame buffer size and not the actual image area we are interested in.

PR #665, opened 2021-07-26
[doc] install: Update new keybindings and command line options

netboy3:
1. Add new keybindings for damage display and frame timing
2. Add command line options for ImGui UI font name and size


netboy3:
As per @gnif, removed damage display, and included a few items that were added to the code since then

PR #685, opened 2021-07-30
[client] config: do not attempt to load non-files as config

quantum5:
Currently, we load /etc/looking-glass-client.ini and/or
~/.config/looking-glass-client.ini as long as they exist, even if they are
not files. We should only load them if they are files.


gnif:
This should also allow symbolic links still.

quantum5:
This works for symlinks to files. `stat` always returns information for the destination.

PR #686, opened 2021-07-30
[client] config: optionally load config from XDG_CONFIG_HOME

quantum5:
We look for the client config in $XDG_CONFIG_HOME/looking-glass/client.ini.
This is done because it's more conventional, and also allows us to add
additional configuration files, e.g. for the host.

We fallback to $HOME/.config as is standard, and then as a last resort use
getpwuid(getuid())->pw_dir. This is also recommended by the getpwuid manpage:

> An application that wants to determine its user's home directory should
> inspect the value of HOME (rather than the value getpwuid(getuid())->pw_dir)
> since this allows the user to modify their notion of "the home directory"
> during a login session.

### DIFF
--- a/client/src/config.c
+++ b/client/src/config.c
@@ -480,7 +480,7 @@ bool config_load(int argc, char * argv[])
       return false;
   }
 
-  // load user's local options
+  // load config from user's home directory
   struct passwd * pw = getpwuid(getuid());
   char * localFile;
   alloc_sprintf(&localFile, "%s/.looking-glass-client.ini", pw->pw_dir);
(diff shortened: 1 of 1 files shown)


client/src/config.c
```diff
@@ -495,6 +495,28 @@ bool config_load(int argc, char * argv[])
...
+  char * dir;
+
+  if ((dir = getenv("XDG_CONFIG_HOME")) != NULL)
+    alloc_sprintf(&xdgFile, "%s/looking-glass/client.ini", dir);
+  else if ((dir = getenv("HOME")) != NULL)
+    alloc_sprintf(&xdgFile, "%s/.config/looking-glass/client.ini", dir);
+  else
+    alloc_sprintf(&xdgFile, "%s/.config/looking-glass/client.ini", pw->pw_dir);
+
+  if (xdgFile && stat(xdgFile, &st) >= 0 && S_ISREG(st.st_mode))
```
gnif, line 509: `S_ISLNK` allowed here too

quantum5: We are using `stat` and not `lstat`, and `stat` returns information for the destination of symlinks. I tested the code as is with symlinks and it loads the file.

PR #690, opened 2021-07-31
[client] ds/imgui: render at screen refresh rate on Wayland

quantum5:
This PR adds the optional display server method `signalNextFrame(LGEvent)`, which takes an LGEvent and signals it when the next frame should be rendered in time for the next vblank. This is currently implemented only for Wayland, and must be invoked before swapping buffers.

This is currently used to render imgui at screen refresh rate.

In theory, this could potentially be used later to implement a better form of vsync for supported display servers. Currently, vsync blocks inside swap until the next vblank, upon which the frame is displayed. This is undesirable as we perform the render early and then waits for the vblank. With `signalNextFrame`, we could render the next frame just in time for vblank.


quantum5:
Currently there appears to be some sort of object leak.

quantum5:
Object leak is fixed.

PR #694, opened 2021-08-01
[client] main/renderer: implement smart vsync mode 

quantum5:
This mode uses `signalNextFrame` to implement vsync without latency if possible, falling back to asking the renderer to use vsync.

We introduce a new parameter `win:vsync`, which enables this logic. In the future, we may make this the default on display servers that support it, i.e. Wayland.

This essentially uses the same `signalNextFrame` logic that we currently use for imgui, but for everything. We automatically enable this mode when overlay is on, preserving the current behaviour.

Currently, this exposes some damage tracking bugs in the EGL renderer, which is why it's not enabled by default.

### DIFF
--- a/client/include/interface/renderer.h
+++ b/client/include/interface/renderer.h
@@ -40,9 +40,8 @@
 
 typedef struct LG_RendererParams
 {
-//  TTF_Font * font;
-//  TTF_Font * alertFont;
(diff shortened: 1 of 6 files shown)


quantum5:
Superseded by #697.

PR #709, opened 2021-08-03
[client] egl: use buffer age extension to render only damaged parts

quantum5:
We avoid rendering any area that has not changed since the buffer was used and also not covered by an overlay.

In order to do this, we refactored the damage mesh generation into a new module, `desktop_rects.c`, which is then used by the desktop module to render its mesh.

As part of this, we simplified and refactored all coordinate space conversions, changing them to use matrices. We are also able to remove the texture copy in the dmabuf path.

### DIFF
--- a/client/renderers/EGL/desktop_rects.c
+++ b/client/renderers/EGL/desktop_rects.c
@@ -219,6 +219,45 @@ struct Rect egl_desktopToScreen(const double matrix[6], const struct FrameDamage
   };
 }
 
+void egl_screenToDesktopMatrix(double matrix[6], int frameWidth, int frameHeight,
+    double translateX, double translateY, double scaleX, double scaleY, LG_RendererRotate rotate,
+    double windowWidth, double windowHeight)
+{
+  double inverted[6] = {0};
+  egl_desktopToScreenMatrix(inverted, frameWidth, frameHeight, translateX, translateY,
+      scaleX, scaleY, rotate, windowWidth, windowHeight);
+
+  double det = inverted[0] * inverted[3] - inverted[1] * inverted[2];
+  matrix[0] =  inverted[3] / det;
+  matrix[1] = -inverted[1] / det;
+  matrix[2] = -inverted[2] / det;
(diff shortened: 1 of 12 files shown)


client/renderers/EGL/desktop_rects.c
```diff
@@ -0,0 +1,267 @@
...
+{
+  *rects = malloc(sizeof(EGL_DesktopRects));
+  if (!*rects)
+  {
+    DEBUG_ERROR("Failed to malloc EGL_DesktopRects");
+    return false;
+  }
+  memset(*rects, 0, sizeof(EGL_DesktopRects));
+
+  glGenVertexArrays(1, &(*rects)->vao);
```
gnif, line 51: I am trying to get this project away from this usage... `(*rects)->`, please use a local, ie:
```
bool egl_desktop_rects_init(EGL_DesktopRects ** rects_, int maxCount)
{
  EGL_DesktopRects * rects = malloc(sizeof(EGL_DesktopRects));
  *rects_ = rects;
  ...
}
```

client/renderers/EGL/desktop_rects.h
```diff
@@ -0,0 +1,54 @@
...
+
+struct DamageRects
+{
+  int count;
+  FrameDamageRect rects[];
+};
+
+typedef struct EGL_DesktopRects EGL_DesktopRects;
+
+bool egl_desktop_rects_init(EGL_DesktopRects ** rects, int maxCount);
```
gnif, line 35: This doesn't need to change now, but I intend to go through and standardize/refactor all the project function names to `camlCase`, it would be helpful if new code was already using this. ie

`eglDesktopRectsInit`

client/renderers/EGL/desktop_rects.c
quantum5: Done.

client/renderers/EGL/desktop_rects.h
quantum5: Done.

PR #712, opened 2021-08-03
[client] main: add graph for time spent rendering

quantum5:
This is the time spent in the on_render function, not the time between
frames. Having this information helps diagnose slowdowns in renderers.


client/src/main.h
```diff
@@ -119,6 +119,7 @@ struct AppState
   uint64_t              lastRenderTime;
   bool                  lastRenderTimeValid;
   RingBuffer            renderTimings;
+  RingBuffer            renderDuration;
```
gnif, line 122: Perhaps these should all be renamed to something more accurate, ie:
```
uploadTiming
renderTiming
renderDuration
```

quantum5: I renamed `frameTimings` to `uploadTimings`. Not sure how the others would be renamed. I think `renderTimings` and `renderDuration` makes sense.

PR #724, opened 2021-08-07
[client] egl/imgui: use imgui for night vision gain configuration

quantum5:


gnif:
Sorry but this is something that is toggled frequently in titles that have a day/night cycle, this needs to stay on a keybind.

quantum5:
Do you want to have it in the config dialog as well or just have the keybind?

gnif:
Doesn't hurt to have both I guess

quantum5:
Done.

PR #725, opened 2021-08-07
[client] egl: invalidate window when toggling night vision

quantum5:
Otherwise, only damaged parts of the window will have night vision.

Note that this is not needed for #724 because the overlay does the invalidation.


quantum5:
Implemented by 12d256c7c86b6225a365543051691b666a1e41f3.

PR #743, opened 2021-08-11
[client] egl: remove textureGather implementation in compat.h

quantum5:
The shader works correctly with #version 320 es and the normal textureGather,
so the problem seems to be the version in compat.h. Since our implementation
needs uInRes[0], we should just move the fallback into ffx_fsr1_easu.frag.


gnif:
nak: I am working on this and the final will use a modified version that's in compat.

PR #753, opened 2021-08-12
[host] dxgi: include correct DXGI headers

quantum5:
The declarations in dxgi_extra.h are not missing, they are in dxgi1_2.h and
dxgi1_5.h, which exist in MinGW-w64 since 2017.


gnif:
They were not in most distibutions of MinGW at the time, thankfully things have changed.

PR #756, opened 2021-08-12
[host] dxgi: correctly count moved rectangles

quantum5:


host/platform/Windows/capture/DXGI/src/dxgi.c
```diff
@@ -776,9 +777,10 @@ static void computeFrameDamage(Texture * tex)
     };
 
     rectToFrameDamageRect(&moveRect->DestinationRect, texDamageRect++);
+    moved += 2;
   }
 
-  tex->damageRectsCount = dirtyRectsCount + moveRectsCount;
+  tex->damageRectsCount = dirtyRectsCount + moved;
```
Xyene, line 783: Shouldn't this be `2 * moveRectsCount` instead? This loop should never break early, because the size of the buffer was already checked.

quantum5: No, because you skip rectangles that haven't actually moved.

quantum5: Renamed variable to be more clear.

Xyene (approved):

PR #769, opened 2021-08-14
[client] add ability to receive keyboard typing in overlays

quantum5:
The motivation for this is to replace the major/minor pixel size hack in the downscale filter configuration, as it is too confusing. This commit replaces that with a text input and a slider.

To do this, we add support for keyboard typing. This is currently implemented for Wayland, though it should be trivial to get X11 working with `libxkbcommon`.


client/src/app.c
```diff
@@ -394,6 +394,24 @@ void app_handleKeyRelease(int sc)
   }
 }
 
+bool app_receiveKeyboardTyping(void)
```
gnif, line 397: `app_isOverlayMode` would be a more appropriate name for this.

quantum5: Fixed.

PR #778, opened 2021-08-15
Use variable-relative sizeof where possible

Xyene:
This eliminates the class of bugs where the size of the variable being assigned doesn't match the amount of memory being allocated.

### DIFF
--- a/client/src/app.c
+++ b/client/src/app.c
@@ -216,7 +216,7 @@ void app_clipboardRequest(const LG_ClipboardReplyFn replyFn, void * opaque)
   if (!g_params.clipboardToLocal)
     return;
 
-  struct CBRequest * cbr = (struct CBRequest *)malloc(sizeof(struct CBRequest));
+  struct CBRequest * cbr = (struct CBRequest *)malloc(sizeof(*cbr));
 
   cbr->type    = g_state.cbType;
   cbr->replyFn = replyFn;
@@ -658,7 +658,7 @@ KeybindHandle app_registerKeybind(int sc, KeybindFn callback, void * opaque, con
     return NULL;
   }
 
-  KeybindHandle handle = (KeybindHandle)malloc(sizeof(struct KeybindHandle));
+  KeybindHandle handle = (KeybindHandle)malloc(sizeof(*handle));
   handle->sc       = sc;
   handle->callback = callback;
   handle->opaque   = opaque;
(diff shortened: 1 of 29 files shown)


client/src/app.c
```diff
@@ -658,7 +658,7 @@ KeybindHandle app_registerKeybind(int sc, KeybindFn callback, void * opaque, con
     return NULL;
   }
 
-  KeybindHandle handle = (KeybindHandle)malloc(sizeof(struct KeybindHandle));
+  KeybindHandle handle = (KeybindHandle)malloc(sizeof(*handle));
```
Xyene, line 661: The original code here made me a little sad, since it leaks the abstraction of what `KeybindHandle` is.

In the original code we had to remember to allocate `sizeof(struct KeybindHandle)` (instead of just `sizeof(KeybindHandle)`), and here we need to allocate `sizeof(*handle)` instead of just `sizeof(handle)`... one has to remember that it's a pointer under the hood.

I propose that separately from this PR, we do away with this typedef (and any others like it).

common/src/stringlist.c
```diff
@@ -32,7 +32,7 @@ struct StringList
 
 StringList stringlist_new(bool owns_strings)
 {
-  StringList sl = malloc(sizeof(struct StringList));
+  StringList sl = malloc(sizeof(*sl));
```
Xyene, line 35: Same here.

client/src/app.c
gnif: Yes, there is much cleanup to be done across the project like this. Thanks for your efforts so far! I have also been slowly working across the project making such cleanups.

gnif (commented):
While I have been at it I have also been converting things to use the following pattern and removing the extra memset.
```
EGL_Splash * this = calloc(sizeof(*this));
if (!this)
{
  // error
  return false;
}

*splash = this;
return true;
```

To standardize the project, all functions should be using `this` as the instance variable after this alteration is made. Not sure if you want to include this as part of this changeset... if not I will keep it on my list of cleanup tasks.

Xyene:
> While I have been at it I have also been converting things to use the following pattern and removing the extra memset.
> 
> ```
> EGL_Splash * this = calloc(1, sizeof(*this));
> if (!this)
> {
>   // error
>   return false;
> }
> 
> *splash = this;
> return true;
> ```
> 
> To standardize the project, all functions should be using `this` as the instance variable after this alteration is made. Not sure if you want to include this as part of this changeset... if not I will keep it on my list of cleanup tasks.

Yep, I noticed the `calloc` inconsistency too... but it looks like @quantum5 has gone and implemented the changes while I was napping, so this PR should be good to go 🙂

PR #800, opened 2021-08-22
[client] egl: allow postprocessing filters to be reordered by dragging

quantum5:
Due to the difficulty of swapping elements with `struct ll`, I created a vector common type (based on C++'s `std::vector`) and switched the storage. There are likely other places that could benefit from this new type (e.g. `EGLImage` storage in `texture_dmabuf.c`), and those will be switched over later.


gnif:
Failing to build on windows.

quantum5:
Fixed.

PR #807, opened 2021-08-28
[client] egl: implement presets

quantum5:
Progress:

* [x] implement option dumping
* [x] implement option setting
* [x] implement presets UI
* [x] make filters save settings
* [x] hook up "Save preset" button
* [x] make `createPreset` save to file
* [x] make filters load settings
* [x] hook up "Load preset" button
* [x] save filter ordering
* [x] load filter ordering
* [x] show modal message dialogs
* [x] implement deletion

This supersedes #805.

### DIFF
--- a/client/renderers/EGL/filter_downscale.c
+++ b/client/renderers/EGL/filter_downscale.c
@@ -417,6 +417,7 @@ static GLuint egl_filterDownscaleRun(EGL_Filter * filter, EGL_Model * model,
 
 EGL_FilterOps egl_filterDownscaleOps =
 {
+  .id           = "downscale",
   .name         = "Downscaler",
   .type         = EGL_FILTER_TYPE_DOWNSCALE,
   .earlyInit    = egl_filterDownscaleEarlyInit,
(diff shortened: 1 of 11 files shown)


client/renderers/EGL/filter_downscale.c
```diff
@@ -77,12 +77,74 @@ static void egl_filterDownscaleEarlyInit(void)
...
+      .module        = "eglFilter",
+      .name          = "downscaleFilter",
+      .description   = "Downscale filter type",
+      .type          = OPTION_TYPE_INT,
+      .value.x_int   = 0
+    },
     { 0 }
   };
 
   option_register(options);
```
gnif, line 118: Should we even be registering these as command-line options at all? This is going to get very messy if we do this as more filters/options are added.

gnif (changes requested):
I don't think we should present all the filter options as command-line options or loadable from `looking-glass-client.ini` Perhaps an option type field is needed to specify if this is a "preset" option or a normal option.

quantum5:
Added a `preset` field to `struct Option` which exempts the option from the help menu. I don't think it's possible to easily prevent such options from being loaded, but no one should try them if they are undocumented.

PR #823, opened 2021-09-01
[doc] Pre B5 Documentation update

netboy3:
This update includes:
- changes to the command line options
- key-bindings
- New section covering the configuration widget
- fixing a few spelling mistakes


doc/install.rst
```diff
@@ -322,6 +321,68 @@ in a special config file $XDG_CONFIG_HOME/looking-glass/imgui.ini.
...
+the post-processing filter stack. Each filter can be expanded to open
+its settings. Filters can also be re-ordered by dragging them up or down.
+Filters application is cumulative, and order is important (e.g. applying CAS
+before FSR might have different results than the reverse). Users are
+encouraged to experiment with the order and parameters to achieve optimal
+results. The currently available filters include:
+
+- Downscaler: Filter for downscaling resolution. It can be used
+  to test how different input resolutions effect the filter stack and
+  find the best stack tuning for the application being used on the guest.
```
quantum5, line 356: The actual goal for the downscaler is to undo the poor upscaling used by games so we can upscale it better with FSR.

doc/install.rst
```diff
@@ -322,6 +321,68 @@ in a special config file $XDG_CONFIG_HOME/looking-glass/imgui.ini.
...
+  application. The CAS filter interface has an adjustable sharpness
+  control.
+
+The filter stack settings and order can be saved to presets so that it
+can be conveniently recalled at a later time. As filter settings are
+usually application specific, multiple presets can be defined for each
+case scenario. To save a preset, click on "Save preset as..." and enter
+a preset name. Presets are recalled by selecting them in the "Preset name"
+pull down. Presets are persistent and are stored on disk at
+``$XDG_CONFIG_HOME/looking-glass/presets``.
```
quantum5, line 379: Is this too much detail? Do we want users to mess with the preset storage?

netboy3: I can add some text that will discourage users from modifying it, but i still think it's a good idea to detail where LG is storing its magic

doc/install.rst
netboy3: I did not know that. Will push an updated section

PR #830, opened 2021-09-02
[client] egl: make filters damage aware

quantum5:
This saves a lot of GPU power for partial updates. Running testufo with
lanczos downscaling and FSR upscaling consumed over 90 W, but with this
commit, consumed only 75 W.


quantum5:
Oops, need to revert changes to the desktop shaders that were not necessary.

quantum5:
Reverted unnecessary changes.

PR #832, opened 2021-09-03
[client] x11: regrab input devices after resuming from sleep

quantum5:
This is meant to fix issues with input devices not working after resuming from sleep, as reported on discord. This should only be merged once confirmed to fix the issue.

To test this, install the bleeding edge host application and apply this PR as a patch to the source code:

```console
$ curl https://patch-diff.githubusercontent.com/raw/gnif/LookingGlass/pull/832.diff | patch -p1
```

Then build the client. Also install the systemd unit:

```console
# cp contrib/looking-glass-client-suspend.service /etc/systemd/system/
# systemctl enable /etc/systemd/system/looking-glass-client-suspend.service
```

Then, suspend and resume the system, and see if the issue is resolved.

### DIFF
--- a/client/displayservers/Wayland/wayland.c
+++ b/client/displayservers/Wayland/wayland.c
@@ -148,6 +148,11 @@ static bool waylandGetProp(LG_DSProperty prop, void * ret)
   return false;
 }
 
+static void waylandSystemChanged(void)
+{
+  // do nothing
+}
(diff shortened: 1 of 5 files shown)


gnif:
Just checking the status of things...
Is this this PR still valid/in-progress etc?

PR #840, opened 2021-09-20
[common] open: don't wait on xdg-open, because it may never exit

Xyene:
This is the case at least on my system opening Firefox; xdg-open stays
around until Firefox exits, freezing LG.

There's nothing we can do if xdg-open were to fail anyway, so let's just
not wait on it.

### DIFF
--- a/common/src/platform/linux/open.c
+++ b/common/src/platform/linux/open.c
@@ -23,7 +23,6 @@
 
 #include <errno.h>
 #include <string.h>
-#include <sys/wait.h>
 #include <unistd.h>
 
 static bool xdgOpen(const char * path)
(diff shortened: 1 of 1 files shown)


quantum5 (approved):

common/src/platform/linux/open.c
```diff
@@ -39,24 +38,8 @@ static bool xdgOpen(const char * path)
...
-    if (WIFEXITED(status) && WEXITSTATUS(status) == 0)
-      return true;
 
-    if (WIFEXITED(status))
-      DEBUG_ERROR("xdg-open exited with code %d", WEXITSTATUS(status));
-    else
-      DEBUG_ERROR("xdg-open exited with signal: %s", strsignal(WTERMSIG(status)));
-    return false;
-  }
+	return true;
```
gnif, line 42: Cosmetics, incorrect indenting

quantum5 (changes requested):
On second thought, this PR is not fine as is because it produces zombie `xdg-open` processes. These must be waited on somehow. We could either create a thread (which would require `lgDetachThread` to be implemented) or fork twice, wait on the immediate child, and let `init` adopt the child.

PR #842, opened 2021-09-21
[host] installer: build 64-bit NSIS installers

quantum5:
The host is a 64-bit application, so requiring 64-bit Windows isn't an
issue. However, not using 32-bit installers has an advantage: we don't
need to require WoW64.

Please test [looking-glass-host-setup.zip](https://github.com/gnif/LookingGlass/files/7202713/looking-glass-host-setup.zip) (requires bleeding edge client).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -127,7 +127,7 @@ jobs:
     - name: Build Windows host installer
       run: |
         cd host\build
-        makensis platform\Windows\installer.nsi
+        makensis -DBUILD_32BIT platform\Windows\installer.nsi
 
   obs:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


garyedwards:
I have tested the zip in this conversation and it installs without issue.  However the Bleeding Edge build on the website based on this commit still does not work with out WoW64.

quantum5:
Strange. I just checked, and the build with commit hash `bc65de59` is most definitely built as 64-bit and should work without WoW64. It might be a caching issue? Can you try hard refreshing the download page and make sure you are downloading `bc65de59` or something newer?

garyedwards:
It works fine, I was using the checksum not the commit hash as the password (even though it is written in bright red) 🤦.  Even B4 now works fine.  Removing WoW64 dep seems to make sense anyway.  Thanks again for your great work.

PR #853, opened 2021-10-03
[client] wayland: implement resizing for libdecor

SytheZN:
# Summary
* Adds autoresize to LG client when running on Wayland with libdecor.
* Fixes an issue where the client would launch pinned to the top left of the display (below GNOME's status bar) as in the screenshot:
image

# Known Issues
* Occasional segfault during resize when moving quickly. Cause appears to be a race condition between resizes from the main render thread and those incoming from the user. Workaround for now is to disable win:forceAspect.

Tested on Fedora 34, GNOME 40.4

### DIFF
--- a/client/displayservers/Wayland/shell_libdecor.c
+++ b/client/displayservers/Wayland/shell_libdecor.c
@@ -51,19 +51,23 @@ static void libdecorHandleError(struct libdecor * context, enum libdecor_error e
 static void libdecorFrameConfigure(struct libdecor_frame * frame,
     struct libdecor_configuration * configuration, void * opaque)
 {
+  if (wlWm.needsResize)
+    return;
+
   int width, height;
-  if (libdecor_configuration_get_content_size(configuration, frame, &width, &height)) {
+  if (libdecor_configuration_get_content_size(configuration, frame, &width, &height)) 
+  {
     wlWm.width = width;
     wlWm.height = height;
+
+    struct libdecor_state * state = libdecor_state_new(wlWm.width, wlWm.height);
+    libdecor_frame_commit(wlWm.libdecorFrame, state, NULL);
+    libdecor_state_free(state);
   }
 
   enum libdecor_window_state windowState;
   if (libdecor_configuration_get_window_state(configuration, &windowState))
     wlWm.fullscreen = windowState & LIBDECOR_WINDOW_STATE_FULLSCREEN;
-
-  struct libdecor_state * state = libdecor_state_new(wlWm.width, wlWm.height);
-  libdecor_frame_commit(frame, state, wlWm.configured ? NULL : configuration);
-  libdecor_state_free(state);
 
   if (wlWm.configured)
   {
@@ -72,7 +76,7 @@ static void libdecorFrameConfigure(struct libdecor_frame * frame,
     app_invalidateWindow(true);
     waylandStopWaitFrame();
   }
-  else
+  else 
     wlWm.configured = true;
 }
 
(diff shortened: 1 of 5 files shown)


client/displayservers/Wayland/shell_libdecor.c
```diff
@@ -51,28 +51,32 @@ static void libdecorHandleError(struct libdecor * context, enum libdecor_error e
...
-
   if (wlWm.configured)
   {
     wlWm.needsResize = true;
     wlWm.resizeSerial = configuration->serial;
     app_invalidateWindow(true);
     waylandStopWaitFrame();
   }
-  else
+  else
```
quantum5, line 79: nit: extra space

client/displayservers/Wayland/shell_libdecor.c
```diff
@@ -149,3 +160,17 @@ void waylandMinimize(void)
...
+  wlWm.height = h;
+
+  struct libdecor_state * state = libdecor_state_new(w, h);
+  libdecor_frame_commit(wlWm.libdecorFrame, state, NULL);
+  libdecor_state_free(state);
+
+  wlWm.needsResize = true;
+  app_invalidateWindow(true);
+  waylandStopWaitFrame();
+}
```
quantum5, line 176: nit: missing `\n`

client/displayservers/Wayland/shell_libdecor.c
```diff
@@ -103,14 +107,19 @@ static void libdecorCallback(uint32_t events, void * opaque)
   libdecor_dispatch(wlWm.libdecor, 0);
 }
 
-bool waylandShellInit(const char * title, bool fullscreen, bool maximize, bool borderless)
+bool waylandShellInit(const char * title, bool fullscreen, bool maximize, bool borderless, bool resizable)
```
quantum5, line 110: The corresponding arguments needs to be added in `shell_xdg.c` as well, even if it's just a stub. In general, the client needs to compile both with `libdecor` support and without.

client/displayservers/Wayland/shell_libdecor.c
```diff
@@ -51,28 +51,32 @@ static void libdecorHandleError(struct libdecor * context, enum libdecor_error e
 static void libdecorFrameConfigure(struct libdecor_frame * frame,
     struct libdecor_configuration * configuration, void * opaque)
 {
+  if (wlWm.needsResize)
+    return;
```
quantum5, line 55: Is this correct? If there are two configure events in a row, the latter should have precedence. We should at least update `wlWm.width` and `wlWm.height` so that the EGL viewport has the correct size when we resize it.

client/displayservers/Wayland/shell_libdecor.c
```diff
@@ -149,3 +160,17 @@ void waylandMinimize(void)
 {
   libdecor_frame_set_minimized(wlWm.libdecorFrame);
 }
+
+void waylandShellResize(int w, int h)
```
quantum5, line 164: A stub needs to be implemented for this in `shell_xdg.c`.

quantum5:
Also the commit message should be something like `[client] wayland: implement resizing for libdecor`.

client/displayservers/Wayland/shell_libdecor.c
SytheZN: I found if the previous resize hadn't been handled yet then when drag was shrinking the window it leads to the renderer attempting to draw on a larger surface than is available. May not be 100%, but I haven't been able to crash it in testing.

client/displayservers/Wayland/shell_libdecor.c
SytheZN: Added

client/displayservers/Wayland/shell_libdecor.c
SytheZN: Added.

client/displayservers/Wayland/shell_libdecor.c
SytheZN: fixed

client/displayservers/Wayland/shell_libdecor.c
SytheZN: Scratch that - seems to work fine without it provided win:forceAspect is disabled.
With forceAspect enabled there's more crashes but that's likely down to the race between frame updates from the cursor and from the timer.

PR #879, opened 2021-10-31
[common] time: fixed time values in nsleep

VMFortress:
Time values of scientific notation replaced with expanded values
and `tv_nsec` simplified with modulo

### DIFF
--- a/common/include/common/time.h
+++ b/common/include/common/time.h
@@ -61,8 +61,8 @@ static inline void nsleep(uint64_t ns)
 #else
   const struct timespec ts =
   {
-    .tv_sec  = ns / 1e9,
-    .tv_nsec = ns - ((ns / 1e9) * 1e9)
+    .tv_sec  = ns / 1000000000,
+    .tv_nsec = ns % 1000000000,
   };
   nanosleep(&ts, NULL);
 #endif


common/include/common/time.h
```diff
@@ -61,8 +61,8 @@ static inline void nsleep(uint64_t ns)
 #else
   const struct timespec ts =
   {
-    .tv_sec  = ns / 1e9,
-    .tv_nsec = ns - ((ns / 1e9) * 1e9)
+    .tv_sec  = ns / (time_t)1000000000L,
```
quantum5, line 64: Why the `(time_t)` cast? If you use the literal without the `L` prefix, it'll be promoted to the same size as the other operand.
```suggestion
    .tv_sec  = ns / 1000000000,
```

common/include/common/time.h
```diff
@@ -61,8 +61,8 @@ static inline void nsleep(uint64_t ns)
...
+    .tv_nsec = ns - ((ns / 1000000000L) * 1000000000L)
```
quantum5, line 65: This generates identical code on all optimization levels ([subtraction](https://godbolt.org/z/v8qs8Pqhh) vs [modulo](https://godbolt.org/z/h8EPh45Wo)), so why don't we just use `%`? 
```suggestion
    .tv_nsec = ns % 1000000000,
```

quantum5 (approved):

PR #885, opened 2021-12-07
[client] wayland: make cursor change work without wl_pointer

quantum5:
This is a bug reported by @Xyene, who should confirm it has fixed his issue.


Xyene (approved):
LGTM!

PR #898, opened 2021-12-25
B5 docs polish pass

JJRcop:
Just looked at the delta from B4, didn't do a full pass of all documentation, so mistakes and uglies from B4 are likely still present.

Thanks @netboy3 and @quantum5 for a majority of the content this time around


doc/install.rst
```diff
@@ -289,112 +289,109 @@ Examples:
...
+FSR might have different results than the reverse). Users are encouraged to 
+experiment with the order and parameters to achieve optimal results. The 
+currently available filters include:
+
+-  Downscaler: Filter for downscaling the host resolution. Can be used to undo 
+   poor upscaling on the VM to better utilize AMD FSR (see below). The filter 
+   has a pixel-size setting that is used to set the effective downscaling ratio,
+   and a configurable interpolation algorithm.
+
+-  AMD FidelityFX Super Resolution (FSR): A spatial upscaling filter that works
```
quantum5, line 380:
```suggestion
-  AMD FidelityFX Super Resolution (FSR): Spatial upscaling filter that works
```
Remove the article to be more consistent with the others.

doc/install.rst
```diff
@@ -289,112 +289,109 @@ Examples:
...
-can be conveniently recalled at a later time. As filter settings are
-usually application specific, multiple presets can be defined for each
-case scenario. To save a preset, click on "Save preset as..." and enter
-a preset name. Presets are recalled by selecting them in the "Preset name"
-pull down. Presets are persistent and are stored on disk at
+The EGL Filters tab contains options for toggling, configuring, and ordering 
+post-processing filters. Each filter can be expanded to open its settings. 
+Filters can also be re-ordered by dragging them up or down. Filters are applied 
+from top to bottom, keep this in mind when ordering them. (applying CAS before 
+FSR might have different results than the reverse). Users are encouraged to
```
quantum5, line 371:
```suggestion
from top to bottom, keep this in mind when ordering them, e.g. applying CAS before 
FSR might have different results than the reverse. Users are encouraged to 
```

doc/obs.rst
```diff
@@ -47,7 +47,7 @@ These instructions are similar to building the
    cmake -DUSER_INSTALL=1 ../
    make
 
-Alternatively if you would like to install the plugin for all users.
+Alternatively, if you would like to install the plugin for all users.
```
quantum5, line 50:
```suggestion
Alternatively, if you would like to install the plugin for all users:
```
Otherwise the sentence looks incomplete, but we haven't been using colons in the documentation that much...

doc/module.rst
```diff
@@ -181,18 +181,19 @@ please create the following XML block in your domain:
 
 .. note::
 
-   - Not using the new syntax above might cause QEMU to abort while starting the VM with
-     "``error: internal error: ... PCI: slot 1 function 0 not available for pcie-root-port, in use by ivshmem-plain``"
+   -  Using an incorrect syntax for your host configuration may cause QEMU to
```
netboy3, line 184: The error doesn't show up just because you have "an incorrect syntax", but specifically if you use the legacy configuration when you should use the new one.

JJRcop: This is what I mean by "incorrect syntax", is there a better way to put it?

PR #904, opened 2021-12-26
[docs] all: Fix mobile layout

JJRcop:


JJRcop:
might as well do the thing I should've done I guess here

PR #907, opened 2021-12-26
[doc] Version no longer gets spellchecked

JJRcop:


doc/conf.py
```diff
@@ -4,8 +4,13 @@
...
 # https://www.sphinx-doc.org/en/master/usage/configuration.html
 
-# -- Path setup --------------------------------------------------------------
-from pathlib import Path
+# -- Release import
+
+import sys, os
+sys.path.append(os.path.dirname(__file__))
+
+
```
quantum5, line 12:
```suggestion
```

doc/conf.py
```diff
@@ -22,30 +27,17 @@
...
 .. |license| replace:: GPLv2
 """
 
+
 # -- General configuration ---------------------------------------------------
 
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
+
```
quantum5, line 40: Is there a reason for these new lines?

doc/lgrelease.py
```diff
@@ -0,0 +1,17 @@
+# -- Path setup --------------------------------------------------------------
+from pathlib import Path
+
+release = None
+
```
quantum5, line 5: This variable is overwritten on all non-error paths.
```suggestion
```

doc/lgspell.py
```diff
@@ -1,5 +1,9 @@
 #!/usr/bin/env python3
 import re
+import sys, os
+sys.path.append(os.path.dirname(__file__))
```
quantum5, line 4: Not necessary here.

gnif (changes requested):
Please fix the white space issues, etc. This PR is rather important as ci builds are failing until it's done.

doc/conf.py
JJRcop: mistake

JJRcop: makes my linter happy but can remove

PR #910, opened 2021-12-27
[misc] Prefer debian package to pip3 when checking spelling

JJRcop:
Swaps the pip3 install command for equivalent Debian package when installing `sphinxcontrib-spelling`

This package is also on Ubuntu since Trusty Tahr


JJRcop:
Apparently this doesn't work!

PR #923, opened 2022-01-10
PipeWire tweaks

spencercw:
My workflow happens to require playback of a lot of short clips, which is pretty glitchy at the moment. There's probably more that could be done in this area, but these changes improve things quite a bit.

Note in particular that without the rate matching change, PipeWire can crash when the quantum size changes. [This has been fixed upstream](https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/1994), but that may take a while to filter through.


client/audiodevs/PipeWire/pipewire.c
```diff
@@ -152,7 +206,9 @@ static void pipewire_playbackStart(int channels, int sampleRate)
   static const struct pw_stream_events events =
   {
     .version = PW_VERSION_STREAM_EVENTS,
-    .process = pipewire_onPlaybackProcess
+    .io_changed = pipewire_onPlaybackIoChanged,
```
gnif, line 209: nit: align `=` with above & below.

gnif, line 209: nit: align `=` with above & below.

client/audiodevs/PipeWire/pipewire.c
```diff
@@ -162,26 +218,39 @@ static void pipewire_playbackStart(int channels, int sampleRate)
...
   pw.playback.channels   = channels;
   pw.playback.sampleRate = sampleRate;
   pw.playback.stride     = sizeof(uint16_t) * channels;
-  pw.playback.buffer     = ringbuffer_new(sampleRate / 10,
+  pw.playback.buffer     = ringbuffer_new(bufferFrames,
       channels * sizeof(uint16_t));
 
+  int maxLatencyFrames = bufferFrames / 2;
+  char * maxLatency;
+  alloc_sprintf(&maxLatency, "%d/%d", maxLatencyFrames, sampleRate);
```
gnif, line 231: do we really need an alloc here? the string is never going to be more then a handful of bytes, ie `snprintf` on a `char[32]`

client/audiodevs/PipeWire/pipewire.c
```diff
@@ -138,9 +192,9 @@ static void pipewire_playbackStopStream(void)
     return;
 
   pw_thread_loop_lock(pw.thread);
-  pw_stream_flush(pw.playback.stream, true);
   pw_stream_destroy(pw.playback.stream);
   pw.playback.stream = NULL;
+  pw.playback.rateMatch = NULL;
```
gnif, line 197: nit: align `=` please

gnif (changes requested):
Awesome work, thanks!
Please add yourself to the `AUTHORS` file too.

gnif:
Sorry one more think I forgot to mention, we are trying to keep all new code to 80 columns, once this is done I see no other reason not to merge this.

spencercw:
No worries, that should be all done now

gnif (approved):

PR #929, opened 2022-01-12
[host] dxgi: fix frame damage method parameters

spencercw:
The buffer input sizes to the `IDXGIOutputDuplication` methods are measured
in bytes. This dramatically increases the number of dirty/move rects that
can be handled.


gnif:
Good catch

PR #935, opened 2022-01-15
[client] Add support for some multimedia keys in capture mode

JJRcop:
These include Volume Up, Volume Down, Mute, Play/Pause, Stop, Next Track, and Previous Track.

I discovered this on accident when messing around with #934.

I tried adding these as ScrLk shortcuts alongside the shortcuts added by that PR, to effectively let you passthrough the keys. So pressing ScrLk+Mute would toggle mute, ScrLk+VolumeUp would press Volume Up, etc.

It wasn't working until I added this code to `kb.c`, but when I did that, the keys themselves were working just fine in capture mode, so I didn't need the ScrLk keybinds after all.

(Later, I added the multimedia keys Play/Pause, Stop, Next Track, and Previous Track)

Added @quantum5 as co-author because his PR is what made me look at this, and the PS2 scancodes were lifted from his PR.

### DIFF
--- a/client/src/kb.c
+++ b/client/src/kb.c
@@ -142,6 +142,9 @@ const uint32_t linux_to_ps2[KEY_MAX] =
   [KEY_F14]              /* = USB 105 */ = 0x00005E,
   [KEY_F15]              /* = USB 106 */ = 0x00005F,
   [KEY_PRINT]            /* = USB  70 */ = 0x00E037,
+  [KEY_MUTE]             /* = USB 127 */ = 0x00E020,
+  [KEY_VOLUMEUP]         /* = USB 128 */ = 0x00E030,
+  [KEY_VOLUMEDOWN]       /* = USB 129 */ = 0x00E02E,
 };
 
 const char * linux_to_str[KEY_MAX] =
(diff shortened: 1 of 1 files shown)


JJRcop:
Pushing re-dated commit to trigger build again
Screenshot_20220115_161243

client/src/kb.c
```diff
@@ -388,6 +402,13 @@ const char * linux_to_display[KEY_MAX] =
   [KEY_F14]              = "F14",
   [KEY_F15]              = "F15",
   [KEY_PRINT]            = "Print",
+  [KEY_MUTE]             = "Mute",
+  [KEY_VOLUMEUP]         = "VolumeUp",
+  [KEY_VOLUMEDOWN]       = "VolumeDown",
+  [KEY_NEXTSONG]         = "NextSong",
+  [KEY_PLAYPAUSE]        = "PlayPause",
+  [KEY_PREVIOUSSONG]     = "PreviousSong",
+  [KEY_STOPCD]           = "StopMedia",
```
JJRcop, line 411: I chose this name instead of STOPCD because it is easier to understand

gnif: Seems fine to me

gnif:
Does this obsolete #934? I'd rather not add extra key-binds for things that are not really needed

quantum5:
I don't think this obsoletes the other PR. I think volume control for just the VM is nice to have and you don't necessarily have these keys on your keyboard or want to enter capture mode just to control the volume.

JJRcop:
It could obsolete #934, if you want to reduce duplicate functionality, but otherwise they both work together just fine. #934 does have a use if capture mode is not used.

quantum5:
Personally I have it set up so that the actual volume keys on my keyboard controls the global volume, bypassing all attempts to capture them at the compositor level. This is so that I could control volume with the screen locked, which also allows me to control the whole system volume in capture mode. In this case, I think I definitely need #934 to control the guest volume.

Also, some people might not have volume control keys on their keyboard, and #934 would also be convenient in that case.

PR #937, opened 2022-01-18
[doc] build: Add source comment for listing deps

JJRcop:
Webedit because I'm super sleepy and don't want to turn on my laptop.

Just a future proofing warning so the rule of the Reqired Dependencies section is shown to prospective future editors and developers.

I don't yet know if this will show up on the rendered html or not, I followed https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#comments so it should be visible in the source only.

### DIFF
--- a/doc/build.rst
+++ b/doc/build.rst
@@ -59,6 +59,13 @@ These required libraries and tools should be installed first.
 Required Dependencies
 ^^^^^^^^^^^^^^^^^^^^^
 
+..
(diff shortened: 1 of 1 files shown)


JJRcop:
I'm being silly by not just converting this to a draft until I make sure. Whups

quantum5 (approved):
Appears that it works.

JJRcop:
Yep

PR #942, opened 2022-01-26
[host] nvfbc: shorten nvfbc_getName() result

quantum5:
To avoid client showing `Using    : NVFBC (NVidia Frame Buffer Capt`.

This happens because the string is truncated to 31 characters to fit in the `char capture[32];` member of `KVMFRRecord_VMInfo`.


8BallBomBom:
Ahahaha beat me to it, literally just noticed 👍🏻

PR #946, opened 2022-01-26
[client] audio: adjust playback speed to match audio device clock

spencercw:
This change is based on the techniques described in [1] and [2].

The input audio stream from Spice is not synchronised to the audio playback device. While the input and output may be both nominally running at 48 kHz, when compared against each other, they will differ by a tiny fraction of a percent. Given enough time (typically on the order of a few hours), this will result in the ring buffer becoming completely full or completely empty. It will stay in this state permanently, periodically resulting in glitches as the buffer repeatedly underruns or overruns.

To address this, adjust the speed of the received data to match the rate at which it is being consumed by the audio device. This will result in a slight pitch shift, but the changes should be small and smooth enough that this is unnoticeable to the user.

The process works roughly as follows:
1. Every time audio data is received from Spice, or consumed by the audio device, sample the current time. These are fed into a pair of delay locked loops to produce smoothed approximations of the two clocks.
2. Compute the difference between the two clocks and compare this against the target latency to produce an error value. This error value will be quite stable during normal operation, but can change quite rapidly due to external factors, particularly at the start of playback. To smooth out any sudden changes in playback speed, which would be noticeable to the user, this value is also filtered through another delay locked loop.
3. Feed this error value into a PI controller to produce a ratio value. This is the target playback speed in order to bring the error value towards zero.
4. Resample the input audio using the computed ratio to apply the speed change. The output of the resampler is what is ultimately inserted into the ring buffer for consumption by the audio device.

Since this process targets a specific latency value, rather than simply trying to rate match the input and output, it also has the effect of 'correcting' latency issues. If a high latency application (such as a media player) is already running, the time between requesting the start of playback and the audio device actually starting to consume samples can be very high, easily in the hundreds of milliseconds. The changes here will automatically adjust the playback speed over the course of a few minutes to bring the latency back down to the target value.

This graph shows the behaviour at the start of playback when a media player is running in the background; notice how the number of buffered samples starts very high then gradually settles at a more reasonable level. It also shows the typical impulse response when the audio device changes the period size. At 400s, playback in the media player was paused, resulting in the period size changing from 2048 to 1024 samples. At 550s, playback was resumed, resulting in the period size reverting to 2048 samples.

rate_change

This graph shows the number of samples in the ring buffer is stable, even after being left running for several hours.

overnight

[1] https://kokkinizita.linuxaudio.org/papers/adapt-resamp.pdf
[2] https://kokkinizita.linuxaudio.org/papers/usingdll.pdf

### DIFF
--- a/client/src/audio.c
+++ b/client/src/audio.c
@@ -26,6 +26,9 @@
 
 #include "dynamic/audiodev.h"
 
+#include <math.h>
+#include <samplerate.h>
+#include <stdalign.h>
 #include <string.h>
 
 typedef enum
@@ -42,6 +45,45 @@ StreamState;
 
 typedef struct
 {
+  int     periodFrames;
+  double  periodSec;
+  int64_t nextTime;
+  int64_t nextPosition;
+  double  b;
+  double  c;
+}
+PlaybackDeviceData;
+
+typedef struct
+{
+  float * framesIn;
+  float * framesOut;
+  int     framesOutSize;
+
+  int     periodFrames;
+  double  periodSec;
+  int64_t nextTime;
+  int64_t nextPosition;
+  double  b;
+  double  c;
+
+  int64_t devLastTime;
+  int64_t devNextTime;
+
+  int64_t devLastPosition;
+  int64_t devNextPosition;
+
+  double  offsetError;
+  double  offsetErrorIntegral;
+
+  double  ratioIntegral;
+
+  SRC_STATE * src;
+}
+PlaybackSpiceData;
+
+typedef struct
+{
   struct LG_AudioDevOps * audioDev;
 
   struct
@@ -50,13 +92,21 @@ typedef struct
     int         volumeChannels;
     uint16_t    volume[8];
     bool        mute;
+    int         channels;
     int         sampleRate;
     int         stride;
     RingBuffer  buffer;
+    RingBuffer  deviceTiming;
 
     LG_Lock     lock;
     RingBuffer  timings;
     GraphHandle graph;
+
+    // These two structs contain data specifically for use in the device and
+    // Spice data threads respectively. Keep them on separate cache lines to
+    // avoid false sharing
+    alignas(64) PlaybackDeviceData deviceData;
+    alignas(64) PlaybackSpiceData  spiceData;
   }
   playback;
 
@@ -74,6 +124,13 @@ typedef struct
 AudioState;
 
 static AudioState audio = { 0 };
+
+typedef struct
+{
+  int64_t nextTime;
+  int64_t nextPosition;
+}
+PlaybackDeviceTick;
 
 static void playbackStopNL(void);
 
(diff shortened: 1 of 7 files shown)


client/src/audio.c
```diff
@@ -142,24 +209,68 @@ static void playbackStopNL(void)
...
+      else
+      {
+        data->nextTime     +=
+          llrint((data->b * error + data->periodSec) * 1.0e9);
+        data->periodSec    += data->c * error;
+        data->nextPosition += frames;
+      }
     }
+
+    PlaybackDeviceTick tick = {
```
gnif, line 261: `{` on next line please.

client/src/audio.c
```diff
@@ -267,14 +388,165 @@ void audio_playbackMute(bool mute)
...
+    if (spiceData->framesIn)
+    {
+      free(spiceData->framesIn);
+      free(spiceData->framesOut);
+    }
+    spiceData->periodFrames  = frames;
+    spiceData->framesIn      = malloc(frames * audio.playback.stride);
+
+    spiceData->framesOutSize = round(frames * 1.1);
+    spiceData->framesOut     =
```
gnif, line 417: check for `malloc` failure and handle appropriately.

common/src/ringbuffer.c
```diff
@@ -101,126 +90,208 @@ int ringbuffer_getLength(const RingBuffer rb)
...
+    }
+
+    if (count > 0)
+    {
+      DEBUG_ASSERT(writeOffset >= 0);
+
+      // We may not be able to write anything if the writer is too far ahead of
+      // the reader
+      uint32_t writeLen = 0;
+      if (writeOffset < rb->length) {
```
gnif, line 153: `{` on new line please

common/src/ringbuffer.c
```diff
@@ -19,79 +19,68 @@
...
 {
-  RingBufferValueFn preOverwriteFn;
-  void *            preOverwriteUdata;
-
-  int     length;
-  size_t  valueSize;
-  LG_Lock lock;
-  int     start, pos, count;
-  char    values[0];
+  uint32_t length;
```
gnif, line 31: why `uint32_t` specifically? Why not just `unsigned int` so the compiler can just select the optimal type for the platform?

gnif (changes requested):

gnif:
Oh, also please update the documentation to include the libsamplerate dependency as well as the github workflow.

gnif:
I am going to branch this as I have a set of improvements/changes also now and will address the changes requested.

gnif:
This PR has been merged

8BallBomBom:
Started having issues after this pr.
Either audio doesn't play at all or it plays for a little while but then randomly goes silent or glitches.
I can see this error in the logs constantly.

```
[E]  54102065716             audio.c:566  | audio_playbackData             | Resampling failed: SRC ratio outside [1/256, 256] range.
```

gnif:
We are aware, please don't flood with reports on this feature yet, it's considered experiemental.

8BallBomBom:
No worries, wasn't sure if it was reported or not, my bad 👍🏻

PR #951, opened 2022-02-07
[host] windows: plumb guest activation request to host

Xyene:
This PR depends on https://github.com/gnif/LookingGlass/pull/950, and implements ferrying guest activation requests to host activation requests.

Currently implemented only for Windows guests.

image

### DIFF
--- a/common/include/common/KVMFR.h
+++ b/common/include/common/KVMFR.h
@@ -139,6 +139,7 @@ typedef struct KVMFRFrame
   uint32_t        damageRectsCount;   // the number of damage rectangles (zero for full-frame damage)
   FrameDamageRect damageRects[KVMFR_MAX_DAMAGE_RECTS];
   bool            blockScreensaver;   // whether the guest has requested to block screensavers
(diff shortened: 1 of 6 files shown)


common/include/common/KVMFR.h
```diff
@@ -139,6 +139,7 @@ typedef struct KVMFRFrame
   uint32_t        damageRectsCount;   // the number of damage rectangles (zero for full-frame damage)
   FrameDamageRect damageRects[KVMFR_MAX_DAMAGE_RECTS];
   bool            blockScreensaver;   // whether the guest has requested to block screensavers
+  bool            requestActivation;  // whether the guest has requested activation since the last frame
```
gnif, line 142: Struct changed, bump KVMFR version

Xyene: Oops, fixed.

gnif: Note: In future we should replace these bools with a bit field so that future enhancements like this do not break protocol compatibility.

PR #952, opened 2022-02-07
[client] x11: Add inline icon

netboy3:
Add a 64x64 LG icon that will show in the WM taskbar when LG is running

Only concern I have here is the license I embedded in `icondata.h`. I've dropped the standard LG license snippet but as this file has imagery data of the LG logo I wasn't sure whether it should be licensed differently.


gnif:
There is no concern with the icon as I commissioned to have it drawn, as such I release it to the project under the project's existing license. I would however prefer not to have this file hard coded, it should be built by CMake at compile time, or better yet, compiled into an object file and then linked as is done for the EGL shaders.

netboy3:
Sure. Let me work on this. I might need another dependency such as libgd to process a PNG. Are you OK with that?

gnif:
Could you not just make the XPM and push it into the repo?

client/displayservers/X11/x11.c
```diff
@@ -23,6 +23,7 @@
 #include "x11.h"
 #include "atoms.h"
 #include "clipboard.h"
+#include "../../../resources/icondata.h"
```
gnif, line 26: Instead of this reverse traversal please make the following change and just `#include "resources/icondata.h"`

```patch
diff --git a/client/CMakeLists.txt b/client/CMakeLists.txt
index ac0bef16..b21278bd 100644
--- a/client/CMakeLists.txt
+++ b/client/CMakeLists.txt
@@ -107,6 +107,7 @@ add_custom_command(
 )
 
 include_directories(
+       ${PROJECT_TOP}
        ${PROJECT_SOURCE_DIR}/include
        ${CMAKE_BINARY_DIR}/include
 )
```

netboy3: all set. i had a feeling you wouldn't like that ugly reverse traversal.

PR #953, opened 2022-02-08
[client] egl: remove extra samplers

hollinsky:
There was an unused sampler in `TextureBuffer`, any time it was bound it used the `base->sampler` instead.

The real point of this, however, is the sampler in the EGL Desktop. If I understand correctly, the desktop doesn't need its own sampler. There is already an identically configured one in the `desktop->texture`.

For some reason, using the texture sampler fixes a black screen issue caused by f0ea882 (where the separate desktop sampler was added) with my GTX 660 using the 470.86 driver. Maybe hitting some limit for how many samplers can be allocated?

### DIFF
--- a/client/renderers/EGL/desktop.c
+++ b/client/renderers/EGL/desktop.c
@@ -57,7 +57,6 @@ struct EGL_Desktop
   EGLDisplay * display;
 
   EGL_Texture          * texture;
-  GLuint                 sampler;
   struct DesktopShader shader;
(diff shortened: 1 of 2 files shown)


gnif:
Changes look sane, I will look a bit deeper into this before committing though. Thanks for identifying the issue, could this perhaps be related to https://forum.level1techs.com/t/looking-glass-b5-01-fsr-issues/181625 ?

PR #960, opened 2022-02-13
[client] audio: fix assertion failure on exit if graph is open

spencercw:
This is a bit intermittent, but can be triggered reliably by putting a short sleep at the beginning of `audio_free`.


gnif:
Not sure this is the right fix here, this method should really not do anything if the application is shutting down.

PR #966, opened 2022-02-27
[client] audio: reduce resampler latency

spencercw:
The best quality resampler has an intrinsic latency of about 3ms, and the processing itself takes another 1-2ms per 10ms block. The faster setting has an intrinsic latency of about 0.4ms, with about 0.04ms processing time. This makes for an overall saving of about 4ms, with negligible loss in quality.

### DIFF
--- a/client/src/audio.c
+++ b/client/src/audio.c
@@ -108,9 +108,9 @@ typedef struct
     RingBuffer  timings;
     GraphHandle graph;
 
-    // These two structs contain data specifically for use in the device and
-    // Spice data threads respectively. Keep them on separate cache lines to
-    // avoid false sharing
+    /* These two structs contain data specifically for use in the device and
+     * Spice data threads respectively. Keep them on separate cache lines to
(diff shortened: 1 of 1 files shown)


client/src/audio.c
```diff
@@ -328,8 +328,7 @@ void audio_playbackStart(int channels, int sampleRate, PSAudioFormat format,
     playbackStop();
 
   int srcError;
-  audio.playback.spiceData.src =
-    src_new(SRC_SINC_BEST_QUALITY, channels, &srcError);
+  audio.playback.spiceData.src = src_new(SRC_SINC_FASTEST, channels, &srcError);
```
gnif, line 331: Line too long, please limit to 80 columns.

spencercw: The line is exactly 80 characters wide. Do you prefer limiting it to 79?

gnif: doh, sorry about that, you're correct.

PR #970, opened 2022-03-04
[client] input: fix confine_pointer argument marshalling

Shootfast:
Under Wayland, if the mouse pointer is disconnected whilst captured
(like say via KVM switch), the waylandWarpPointer code will be called
but the pointer will be NULL. This results in the cryptic message:

error marshalling arguments for confine_pointer (signature noo?ou): null value passed for arg 2
Error marshalling request: Invalid argument

This patch adds a check on the wlWm.pointer pointer before attempting
to warp the pointer, and avoids the crash.


gnif:
Thanks, looks good but before we can commit this you need to add yourself to the `AUTHORS` file.

Shootfast:
Ah thanks, I didn't know if it was worth it for such a tiny change. I've added myself to the `AUTHORS` file.

gnif:
Sorry one more thing I missed. It has not yet been announced officially but we are starting to enforce 80 column line limits. If you could adjust your change it would be appreciated.

Shootfast:
Sure thing. Is this what you had in mind?

gnif:
A single line wrap would have been sufficient, but this is fine. Thanks!

PR #971, opened 2022-03-06
[host] dxgi: add microsecond-scale sleep capability to d3d12

VMFortress:
The usleep() call lets d3d12 match the sleep calls of d3d11 while
maintaining the current millisecond-scale sleep interface in the
configuration file.

### DIFF
--- a/host/platform/Windows/capture/DXGI/src/d3d12.c
+++ b/host/platform/Windows/capture/DXGI/src/d3d12.c
@@ -23,6 +23,7 @@
 #include <assert.h>
 #include <d3d12.h>
 #include <d3d12sdklayers.h>
+#include "common/time.h"
 #include "common/debug.h"
 #include "common/option.h"
 #include "common/windebug.h"
@@ -44,7 +45,7 @@ struct D3D12Texture
 
 struct D3D12Backend
 {
-  int                   copySleep;
+  float                 copySleep;
   ID3D12Device        * device;
   ID3D12InfoQueue1    * debugInfoQueue;
   ID3D12CommandQueue  * commandQueue;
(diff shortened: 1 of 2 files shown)


gnif:
We opted to not use `usleep` here due to system resolution issues. If you truly need sub ms accuracy please use `nanosleep`.

host/platform/Windows/capture/DXGI/src/d3d12.c
```diff
@@ -120,8 +121,8 @@ static bool d3d12_create(struct DXGIInterface * intf)
     return false;
   }
 
-  this->copySleep = option_get_int("dxgi", "d3d12CopySleep");
-  DEBUG_INFO("Sleep before copy : %d ms", this->copySleep);
+  this->copySleep = option_get_float("dxgi", "d3d12CopySleep");
```
quantum5, line 124: `d3d12CopySleep` is declared as `OPTION_TYPE_INT`, so this would just generate garbage.

host/platform/Windows/capture/DXGI/src/d3d12.c
```diff
@@ -346,7 +347,7 @@ static bool d3d12_copyFrame(Texture * parent, ID3D11Texture2D * src)
   HRESULT status;
 
   if (this->copySleep > 0)
-    Sleep(this->copySleep);
+    usleep((uint64_t)(this->copySleep * 1000));
```
quantum5, line 350: We have a custom `nsleep` handler that uses `NtDelayExecution` and is supposed to be more accurate.

VMFortress:
Updated to `nsleep()`. Apologies as I was using `d3d11.c` as reference but I think we've discussed some of this before.

Also, I fixed the type define in `dxgi.c`.

quantum5 (approved):
LGTM

PR #973, opened 2022-03-07
[host] app: catch failed allocation of mouse pointer memory

VMFortress:


host/src/app.c
```diff
@@ -390,6 +390,10 @@ static bool captureStop(void)
 bool captureGetPointerBuffer(void ** data, uint32_t * size)
 {
   PLGMPMemory mem = app.pointerShapeMemory[app.pointerShapeIndex];
+  if(!mem)
+  {
```
gnif, line 394: formatting, no need for braces for a single if statement.
```c
if (x)
  something();
```

VMFortress: Mixed up work standards and LG standards. Should be resolved!

VMFortress:
Closing as crash should be expected if `captureGetPointerBuffer()` is called before the completion of init.

PR #980, opened 2022-04-09
[client] audio/pw: pw_stream_get_time() deprecated

netboy3:
pw_stream_get_time() is deprecated in PipeWire 0.3.50. Use pw_stream_get_time_n() instead based on PipeWire
library version.

As of this writing the PipeWire master branch is still at 0.3.49, but the deprecation commit has been merged about a week ago, so this patch will not solve build issues until master is updated to 0.3.50. To build from PW master for now we can turn off Werror for deprecations `CFLAGS='-Wno-error=deprecated-declarations' cmake ..`

### DIFF
--- a/client/audiodevs/PipeWire/pipewire.c
+++ b/client/audiodevs/PipeWire/pipewire.c
@@ -368,7 +368,11 @@ static size_t pipewire_playbackLatency(void)
   struct pw_time time = { 0 };
 
   pw_thread_loop_lock(pw.thread);
+#if PW_CHECK_VERSION(0, 3, 50)
+  if (pw_stream_get_time_n(pw.playback.stream, &time, sizeof(time)) < 0)
+#else
   if (pw_stream_get_time(pw.playback.stream, &time) < 0)
+#endif
     DEBUG_ERROR("pw_stream_get_time failed");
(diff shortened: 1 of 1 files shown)


client/audiodevs/PipeWire/pipewire.c
```diff
@@ -368,7 +368,11 @@ static size_t pipewire_playbackLatency(void)
   struct pw_time time = { 0 };
 
   pw_thread_loop_lock(pw.thread);
+#if PW_MAJOR >= 0 && PW_MINOR >= 3 && PW_MICRO >= 50
```
quantum5, line 371: This logic is wrong. For example, PipeWire 1.0.0 will fail this check and fall back to the legacy code.

quantum5 (changes requested):

client/audiodevs/PipeWire/pipewire.c
quantum5: You can use the `PW_CHECK_VERSION` macro for this btw.

quantum5 (approved):

client/audiodevs/PipeWire/pipewire.c
netboy3: You're right. fixed.

PR #982, opened 2022-04-16
[doc] Move Usage section in Install to own page

JJRcop:
Also includes a few miscellaneous strawberry flavoured formatting changes sprinkled on top to taste.


doc/usage.rst
```diff
@@ -0,0 +1,400 @@
...
+.. _client_config_options_file:
+
+Configuration Files
+~~~~~~~~~~~~~~~~~~~
+
+By default, Looking Glass will load config files from
+the following locations:
+
+-  /etc/looking-glass-client.ini
+-  ~/.looking-glass-client.ini
```
JJRcop, line 97: Were these deprecated? I heard something about that from someone on discord

gnif: Yes, they still work but we want people to move across to using the XDG_CONFIG path.

JJRcop: Alright, should I add a notice here?

gnif: I know I said otherwise earlier, but on 2nd thoughts we will keep it, let people decide

PR #983, opened 2022-04-16
Small FAQ Tekst Change and VirtIO Keyboard + Mouse Solution

Stetsed:
Added adding a virtio keyboard and mouse as the first solution to try to fix jumpy slow and laggy mouse, and then if that doesn't work suggesting to try and install the SPICE guest tools.


gnif:
1) Please discuss changes like this with @JJRcop on discord who manages the documentation.
2) Commit message format is wrong, please see other commits for example
3) Limit line length to 80 columns.

PR #987, opened 2022-05-04
[client] x11CheckEWMHSupport() causes seg fault if `_NET_SUPPORTING_WM_CHECK` is empty

noonebtw:
this caused a crash when running under a window manager that doesn't provide a
valid window in `_NET_SUPPORTING_WM_CHECK`


gnif:
Looks good but a few minor things:

1. Commit message format needs to follow the project standard, for this it would be:
  `[client] x11: add null check on data returned by XGetWindowProperty`
2. You need to add yourself to the `AUTHORS` granting this project ownership of your contribution.

gnif:
Thanks for the update but you have not fixed the commit, but only renamed this pull request.

gnif:
Due to the simplicity of this fix and the confusion over the commit message I have committed this fix directly.

PR #988, opened 2022-05-13
Texture compression and transcoding support

AIlucky:
This PR adds support for GPU-native texture compression, as well as 8bpc, 4-to-3 channel transcodes, such as BGRA to RGB. These methods have been introduced in our libttc library which we have included in the submodules. Currently, these changes are limited to CPU compression on the DXGI backend, although we do plan to add support for more capture backends in the future.

We have also moved data which used to be in framebuffer.c into framebuffer.h, to allow for external programs, such as lgproxy which we use for network transmission of Looking Glass data to build directly against the upstream Looking Glass project rather than maintaining a separate fork.

Texture transcoding, especially for BGRA to RGB, may improve performance on bandwidth-constrained GPUs, especially those connected through chipset lanes. These results were gathered using the throughput metrics available via `nvidia-smi`.

Resolution: 1920 x 1080, 30Hz
ETC2 46,000 KB/s
ETC2 (EAC) 69,567 KB/s
DXT5 87,333 KB/s
DXT1 49,250 KB/s
RGB 228,977 KB/s
Raw BGRA 306,950 KB/s

If there are any issues, please let us know. Thanks

Co-authored-by: Kittidech Ditsuwan @poiler22 <bright_ra2@hotmail.com>
Signed-off-by: Tim Dettmar @beanfacts <timd@bordery.net>


gnif:
I am sorry but  as it stands this PR will not be merged for the following reasons:

- Numerous code formatting issues, does not conform to the current coding styles used by this project.
- Numerous white-space issues and cosmetic changes that are not even required (ie: https://github.com/gnif/LookingGlass/pull/988/files#diff-6f4207bbb71067c64a15502fb56e701040b23a0422d8459a3ef92dde685eb641R541).
- Everything has been put into a single commit making this very hard to audit, each major change-set should be split into it's own commit.
- Failure to use an option parser as mentioned in discord, you have implemented a validator but then proceed to perform `_strnicmp` instead of using the parser to convert to an enumeration for later use.
- Functions such as `_strnicmp` are Microsoft defined and not part of the ISO C standard.
- The inability to compile with these features disabled entirely, remember LG's primary goals are lossless and ultra low latency, the only feature here that is actually lossless is the RGB24 support.
- The use of an external library for these conversions makes performing them using accelerated methods with the GPU very difficult due to the need to access and use the OpenGL context.

The last point is the most major one, anything that performs such conversions needs to be done in a way that can also be done in the GPU and handled by the renderer itself. LG is already on the verge of being unusable for many at 4k resolutions due to memory bandwidth limitations, and many people are using LG on laptops or systems with low core counts (most common we see is 6 cores).

Your test metrics are using 1080p @ 30Hz, which is well below our minimum target of 1080p @ 60Hz. On a general enthusiast desktop PC (6 core, 12 thread), how does this perform at our minimum target? If it can't maintain 1080p @ 60Hz minimum without frame drops or micro-stutters there is still much work to do before we could include it.

gnif:
After to the discussions on Discord I am closing this issue for now. Until a elegant solution of offloading these transformations to the GPU is found and developed this will not be merged into this project.

PR #990, opened 2022-05-16
Migrate framebuffer metadata to header

matthewjmc:
As discussed on Discord a use case for us is to allow for framebuffer data to be sent, as soon as it is available, over the network and update it in real-time via RDMA. This requires us to know the exact framebuffer write pointer location. See below for details:

lgproxy-ptr

This pull request exposes framebuffer metadata so it may be used by external code which interfaces with the Looking Glass framebuffer on a low level.


common/include/common/framebuffer.h
```diff
@@ -24,9 +24,23 @@
...
+#define FB_WP_TYPE      atomic_uint_least32_t
+#define FB_WP_SIZE      sizeof(FB_WP_TYPE)
+
+struct stFrameBuffer
+{
+  FB_WP_TYPE wp;
+  uint8_t    data[0];
+};
+
+static const size_t FrameBufferStructSize = sizeof(FrameBuffer);
```
quantum5, line 42: At this point we might as well get rid of this and just inline `sizeof(FrameBuffer)`, since `FrameBufferStructSize` only existed as a hack to avoid exposing `struct stFrameBuffer`.

common/include/common/framebuffer.h
```diff
@@ -24,9 +24,23 @@
...
+#include <stdatomic.h>
 
 typedef struct stFrameBuffer FrameBuffer;
 
+#define FB_CHUNK_SIZE   1048576 // 1MB
+#define FB_SPIN_LIMIT   10000   // 10ms
+#define FB_WP_TYPE      atomic_uint_least32_t
+#define FB_WP_SIZE      sizeof(FB_WP_TYPE)
+
+struct stFrameBuffer
```
quantum5, line 36: Might be worth merging this with the `typedef struct stFrameBuffer FrameBuffer;` above.

gnif (changes requested):
+1 on @quantum5's comments

matthewjmc:
Do you prefer we just #define FrameBufferStructSize sizeof(FrameBuffer) or modify all values in source code?

gnif:
> Do you prefer we just #define FrameBufferStructSize sizeof(FrameBuffer) or modify all values in source code?

Update the source to use `sizeof` please

matthewjmc:
Please check the new changes. Thanks

PR #999, opened 2022-05-30
[module] fix compile on 5.18

Leo1998:
module compilation fails on kernel 5.18 because dev_pagemap, devm_memremap_pages and devm_memunmap_pages are not included anymore. memremap.h must be explicitly included to fix this.

Tested on Arch kernel: 5.18.0-arch1-1


gnif:
Hi,

Thanks for this however due to project policy you also need to add yourself to the `AUTHORS` file before we can merge this PR.

PR #1000, opened 2022-06-02
[client] wayland: fix infinite resize loop

quantum5:
The intention has been to keep `wlWm.needsResize` true when skipping resize, but I accidentally negated the value.


i418c:
I pulled this in and tested it, and it still works ok.

PR #1001, opened 2022-06-02
[client] keybind: add ScrollLock+C to toggle microphone defaults

quantum5:
This makes it possible to change the default action taken the next time an application tries to open the microphone without restarting the client.

### DIFF
--- a/client/src/keybind.c
+++ b/client/src/keybind.c
@@ -139,6 +139,27 @@ void keybind_commonRegister(void)
       "Toggle overlay");
 }
 
+#if ENABLE_AUDIO
+static void bind_toggleMicDefault(int sc, void * opaque)
+{
+  g_state.micDefaultState = (g_state.micDefaultState + 1) % MIC_DEFAULT_MAX;
+
+  switch (g_state.micDefaultState)
+  {
+    case MIC_DEFAULT_PROMPT:
+      app_alert(LG_ALERT_INFO, "Microphone access will prompt");
+      break;
+
+    case MIC_DEFAULT_ALLOW:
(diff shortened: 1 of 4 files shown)


client/src/keybind.c
```diff
@@ -168,6 +189,8 @@ void keybind_spiceRegister(void)
 #if ENABLE_AUDIO
     if (audio_supportsRecord())
     {
+      app_registerKeybind(KEY_A, bind_toggleMicDefault, NULL,
+          "Toggle audio recording default");
```
JJRcop, line 193:
```suggestion
          "Cycle audio recording default");
```
I think this is more accurate.

quantum5: Fixed.

client/src/main.h
```diff
@@ -50,6 +50,7 @@ enum MicDefaultState {
   MIC_DEFAULT_ALLOW,
   MIC_DEFAULT_DENY
 };
+#define MIC_DEFAULT_MAX (MIC_DEFAULT_DENY + 1)
```
gnif, line 53: This doesn't follow the format we use for other enumerations, this should be an entry in the enum itself

quantum5: We should change the format. Putting `MIC_DEFAULT_MAX` will require an extra `default` branch when switching or `gcc` will issue a warning saying that not all enum elements are matched. I don't think it's a good idea to keep wrestling with the compiler like that. We already do this with [`LG_ROTATE_MAX`](https://github.com/gnif/LookingGlass/blob/fba7c80b2f7fd134a956aaab4d88b8e7b1b4b294/client/include/interface/renderer.h#L68-L70).

PR #1008, opened 2022-06-29
[obs] Add a setting that toggles showing the mouse cursor

0xdc:
Give OBS plugin users the option of hiding or showing the mouse cursor, since Looking Glass renders it client-side.

Note: for new or upgrading users, the setting will default to having the mouse cursor hidden.

PR is based against B5 branch, but should also apply cleanly to master.


gnif:
Looks good but please adjust your commit message to wrap at 80 columns.

gnif (approved):

PR #1011, opened 2022-07-10
Install .desktop file and additional icon sizes for client

snowkat:
This installs "Looking Glass Client" to the System menu category along with various sizes of the icon.


JJRcop:
Should these images instead be generated using a source image and maybe something like imagemagick during make install?

gnif:
This is also the wrong icon, it needs updating to use https://github.com/gnif/LookingGlass/blob/master/resources/lg-logo.svg

snowkat:
> Should these images instead be generated using a source image and maybe something like imagemagick during make install?

I converted them on my end to avoid having a build-time dependency on imagemagick. If dynamically generating them is preferred, I can modify the functionality for that.

> This is also the wrong icon, it needs updating to use https://github.com/gnif/LookingGlass/blob/master/resources/lg-logo.svg

Good to know. Since SVG support is optional (per https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html#directory_layout), would it be safe to use https://github.com/gnif/LookingGlass/blob/master/resources/lg-logo.png in addition for systems which don't support SVG icons?

gnif:
> I converted them on my end to avoid having a build-time dependency on imagemagick. If dynamically generating them is preferred, I can modify the functionality for that.

I don't think we need to do such a think, just simply re-generate them and commit them to the repository.

> would it be safe to use https://github.com/gnif/LookingGlass/blob/master/resources/lg-logo.png in addition for systems which don't support SVG icons?

Yes provided it is updated first :)

gnif:
Closing in favour of #1142

PR #1016, opened 2022-08-08
[obs] Allow the client to auto-recover

esibun:
Finally decided to submit this for feedback.

Previously, if the client's subscription to the frame buffer becomes invalid for any reason, the video feed in OBS will freeze until the user goes in and changes any of the settings.  This commit allows the plugin to automatically attempt to recover.

This also brings the OBS client in line with the behavior of the standalone client which will currently automatically restart in the same case.

You can see this in action by restarting the LG service while OBS is running - the client should resubscribe to the frame and pointer data automatically.

This won't solve underlying issues not caused by LG (driver crashes, etc.), just makes them less annoying to deal with while streaming.


gnif:
Looks good but instead of duplicating code please create a function for `pthread_create` and `pthread_join`

esibun:
Brought the pthread calls into two functions to avoid duplicate code.

PR #1018, opened 2022-09-15
X11 mouse is invisible even if hideCursor is false

harikattar:
x11Init assumes you're going to be hiding the mouse, so defines the cursor unconditionally.

FWIW the 'dot' cursor is always invisible on my system if it works at all.

I don't see a good way to fix this.  The other option is to take the XDefineCursor call out of x11Init() entirely, and leave it up to the rest of the code if .SetPointer() should be used.

### DIFF
--- a/client/displayservers/X11/x11.c
+++ b/client/displayservers/X11/x11.c
@@ -58,6 +58,7 @@
 #define _NET_WM_STATE_TOGGLE 2
 
 struct X11DSState x11;
+struct LG_DisplayServerOps LGDS_X11;
 
 struct MwmHints
 {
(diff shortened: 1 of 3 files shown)


harikattar:
FWIW I don't think this should be merged. It's crude, it exports more information across module boundaries and it's not obvious from looking at the source that LGDS_X11.setPointer() might not point to x11SetPointer(). It's mostly here to demonstrate the issue I'm having and a way to demo the desired behavior.

Fundamentally the issue here is that if we call setPointer() at any point we've taken charge of the pointer from the window manager. I think setting input:hideMouse=no (or a similar flag) should just leave it alone. There's only 7 callsites and a number of them are in the same function.

gnif:
This PR is not applicable anymore as we have changed the way the dot cursor is rendered.

PR #1031, opened 2022-09-26
Remove non-prototype function declarations

thrifty-txt:
With -Wstrict-prototypes on non-protyped functions are deprecated and functions must include a void parameter if they do not take parameters.


gnif:
Thanks.

1) Add yourself to the `AUTHORS` file
2) Use the project commit message format, for example `[client] remove non-prototype function declarations`
3) We do not allow altering of the third party source such as nanosvg as it makes maintaining it difficult. Please submit your fix upstream to them and we will then update the library here.

thrifty-txt:
Done. Thanks for the heads-up.

thrifty-txt:
Upstream PR can be found here https://github.com/memononen/nanosvg/pull/229

thrifty-txt:
Upstream accepted the patch.

PR #1037, opened 2022-10-14
[client] corrected mistaken constant name

suphan-barzani:
I was building the client in my Debian system, and this error came up; I think that the name of the constant was wrong. This change worked for me. I hope it helps.


netboy3:
PW_KEY_CONFIG_NAME was declared in Pipewire more than a year ago. Please update your Pipewire version.

gnif:
`PW_CORE_NAME` is an entirely different property, @netboy3 is correct in that your pipewire is too old.

PR #1042, opened 2022-11-09
Draw a tooltip if the user mouses over the spice icon

babbaj:


client/src/overlay/status.c
```diff
@@ -99,27 +100,52 @@ static int status_render(void * udata, bool interactive, struct Rect * windowRec
...
       (ImVec2){ 1, 1 },
       0xFFFFFFFF);
 
+    // draw a helpful little tooltip if the user tries to mouse over the spice icon
+    if (i == LG_USER_STATUS_SPICE && g_cursor.valid) {
+      const double x = g_state.io->MousePos.x;
+      const double y = g_state.io->MousePos.y;
+      if (x >= bottomLeft.x && y <= bottomLeft.y && x <= topRight.x && y >= topRight.y) {
+        igBeginTooltipEx(ImGuiTooltipFlags_None, ImGuiWindowFlags_None);
+        igText("Using looking glass through SPICE emulation, there is likely a problem with your setup");
```
jonpas, line 126:
```suggestion
        igText("Using Looking Glass through SPICE emulation, there is likely a problem with your setup");
```

gnif:
I am sorry but this breaks interaction with anything that might be under the icon. It also is hard coding a lot of things that ideally should be loaded dynamically. And finally, changes like this are not being accepted at this time until we are back to our regular release cycle.

babbaj:
This shouldn't break interaction any more than the icon already does

PR #1043, opened 2022-11-09
[ci] case insensitive grep for AUTHORS file

babbaj:
fails for me because my github username is lowercase but I put an uppercase B in the AUTHORS file


gnif:
This is intentional, we need the username to match up correctly..

PR #1058, opened 2023-01-28
Added information about NVFBC deprecation

marmeladapk:


gnif:
Until NvFBC is removed from the NVIDIA driver we see no need to include this. NvFBC is far faster and superior to the junk DXGI Desktop Duplication API and people should be using it if they can.

PR #1071, opened 2023-05-01
[common] option: rtrim boolean parser

netboy3:
User on Discord complained that they are getting different results from LG when specifying `egl:vsync=yes` on the command-line and in the config file. Doing cut-and-paste from their posted config, showed that they had a trailing space after the option, which caused the parser to fail.
This PR adds rtrim the value received by the options boolean parser so trailing white space in the config file will not fail the comparison. The rtrim code shamelessly copied from another section of options.c

### DIFF
--- a/common/src/option.c
+++ b/common/src/option.c
@@ -63,11 +63,20 @@ static bool int_parser(struct Option * opt, const char * str)
 
 static bool bool_parser(struct Option * opt, const char * str)
 {
-  opt->value.x_bool =
-    strcasecmp(str, "1"   ) == 0 ||
-    strcasecmp(str, "on"  ) == 0 ||
-    strcasecmp(str, "yes" ) == 0 ||
-    strcasecmp(str, "true") == 0;
+  char val[5];
+  size_t len = strlen(str);
+  while (len > 1 && isspace(str[len-1]))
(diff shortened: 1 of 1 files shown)


common/src/option.c
```diff
@@ -63,11 +63,19 @@ static bool int_parser(struct Option * opt, const char * str)
 
 static bool bool_parser(struct Option * opt, const char * str)
 {
+  //rtrim
+  char * val = strdup(str);
+  size_t len = strlen(val);
+  while (len > 1 && isspace(val[len-1]))
+    --len;
+  val[len] = '\0';
```
quantum5, line 71: Can we just use a stack buffer with 5 characters, then fails for longer strings and then `memcpy`?

netboy3: How will that detect "1_" (with "_" being space)?

quantum5: You use the same logic here to compute the "actual" length.

quantum5 (approved):

gnif:
Sorry but I do not agree with how this is being fixed, the bug isn't the bool parser itself. Options really never have a reason to have a trailing space for any type and this should be fixed in the parser itself.

PR #1073, opened 2023-05-02
[common] options: Fix regression in option parsing logic

netboy3:
Oops on command line parser logic argv counter


gnif:
Thanks, nice catch!

PR #1082, opened 2023-06-12
Update udev rule in module.rst

f-fouad:
Udev rules accept UIDs. Use that to have a functional snippet by default and prompt to change if necessary.


netboy3:
I don't see the advantage in this change. Assuming every system will have its active user as the first created user and that user's UID is 1000 is not necessarily correct. UIDs are more confusing to users than usernames. A better choice, in my opinion, to make a more KISS solution for this is to remove the `OWNER` parameter and add a `TAG+="uaccess"` parameter. This will use the systemd currently logged-in user to apply dynamic ACL to the udev node.

PR #1085, opened 2023-07-23
Update libdecor source link in FAQ and package name to dev package

JJRcop:
Update link from https://gitlab.gnome.org/jadahl/libdecor to https://gitlab.freedesktop.org/libdecor/libdecor
Change package name from `libdecor` to `libdecor-0-dev` to help those that may install the non-devel dependency.

### DIFF
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -103,7 +103,7 @@ which will allow you to move the window and resize it.
    alternatives if you require stable operation.
 
 .. _standard protocol: https://wayland.app/protocols/xdg-decoration-unstable-v1
-.. _libdecor: https://gitlab.gnome.org/jadahl/libdecor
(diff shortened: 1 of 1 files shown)


netboy3:
> Also explicitly include `libdecor-0-0` to eliminate reliance on the Debian dependency relationship between them.

Why? I thought we only list the necessary packages and not include their dependency path. All library dev packages will always have a dependency on their runtime shared library package. If we decide to change the process to "eliminate reliance on the Debian dependency relationship", we have to add the non-dev packages throughout the documentation as they are all missing.

JJRcop:
> Why? I thought we only list the necessary packages and not include their dependency path. All library dev packages will always have a dependency on their runtime shared library package.

Of course! It slipped my mind at the time. Maybe not a great thing to happen to the documentation guy, but thanks for the reminder.

PR #1087, opened 2023-08-16
[client] wayland: support back/forward buttons on some mouses

MakiseKurisu:
Despite their names, SPICE_MOUSE_BUTTON_SIDE and SPICE_MOUSE_BUTTON_EXTRA are actually treated by Windows as BTN_BACK and BTN_FORWARD, with Edge browser and Heroes of the Storm both recognize them and label them correctly.

Add the missing mapping for BTN_BACK and BTN_FORWARD.

This is hinted by [1].

1: https://github.com/rust-windowing/winit/issues/1914#issuecomment-826087457

### DIFF
--- a/client/displayservers/Wayland/input.c
+++ b/client/displayservers/Wayland/input.c
@@ -110,8 +110,10 @@ static int mapWaylandToSpiceButton(uint32_t button)
     case BTN_RIGHT:
       return 3;  // SPICE_MOUSE_BUTTON_RIGHT
     case BTN_SIDE:
+    case BTN_BACK:
       return 6;  // SPICE_MOUSE_BUTTON_SIDE
     case BTN_EXTRA:
+    case BTN_FORWARD:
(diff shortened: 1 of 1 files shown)


gnif:
Sorry but this doesn't make much sense, we are already aware of how windows treats these buttons and we have many Wayland users already making use of the side buttons on their mice without issue.

If you have hardware sending different event codes, this is either an upstream bug or a bug with your hardware. If we implement the above change it may start breaking behaviour on other mice that use these event codes for other different buttons.

MakiseKurisu:
After some testing I'd say you are probably right. I use Piper and I was having issue with side buttons that I had to create a different mouse profile to bind those buttons to keyboard keys to get registered inside LG host. I did full system upgrade before working on this PR to install some dependencies and now the issue resolves itself, and my test result is likely due to this instead of my PR. I retested with the latest stable LG and it is working as intended.

PR #1095, opened 2023-10-30
[doc] html_unescape: Create html.unescape extension and use it for no-break space

JJRcop:
Fixes no-break space by adding a new html_unescape extension to the docs, which runs Python's `html.unescape` on source files before Sphinx renders them.

This lets us use HTML references like `&nbsp;` in the docs, which get turned into the real characters as Sphinx is reading the files to render them. The source files are not affected, this only happens when rendering.

I have also published this extension in my own name under a different license (the same one Sphinx uses) for others to use:
https://github.com/JJRcop/sphinxcontrib-html_unescape

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -37,6 +37,7 @@ rst_prolog = """
 
 extensions = [
     'sphinx_rtd_theme',
+    'html_unescape',
 ]
 
 try:
@@ -63,6 +64,14 @@ exclude_patterns = ['_build', 'Thumbs.db', '.DS_Store']
 
 # Explicitly state master_doc instead of relying on default
 master_doc = 'index'
+
+# Documents to exclude from html_unescape
+# For example ['index', 'credits']
+html_unescape_exclude = []
+
+# If not empty, html_unescape will ignore any documents not in the list
+# For example ['install', 'usage']
+html_unescape_onlyinclude = []
 
 # -- Options for HTML output -------------------------------------------------
 
--- a/doc/html_unescape.py
+++ b/doc/html_unescape.py
@@ -1,0 +1,58 @@
+# Copyright 2023 The Looking Glass Authors
+# https://looking-glass.io
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License as published by the Free
+# Software Foundation; either version 2 of the License, or (at your option)
+# any later version.
+#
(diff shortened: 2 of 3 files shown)


gnif:
Is there a better way to do this, such as `&nbsp;`?

JJRcop:
> Is there a better way to do this, such as `&nbsp;`?

The docutils (sphinx sits on top of this) FAQ page seems to recommend using the literal character rather than escaping it.

### [How can I represent esoteric characters (e.g. character entities) in a document?](https://docutils.sourceforge.io/FAQ.html#how-can-i-represent-esoteric-characters-e-g-character-entities-in-a-document)
> For example, say you want an em-dash (XML character entity &mdash;, Unicode character U+2014) in your document: use a real em-dash. Insert literal characters (e.g. type a real em-dash) into your input file, using whatever encoding suits your application, and tell Docutils the input encoding. Docutils uses Unicode internally, so the em-dash character is U+2014 internally.
> […]
> ReStructuredText has no character entity subsystem; it doesn't know anything about XML character entities. To Docutils, "&mdash;" in input text is 7 discrete characters; no interpretation happens. When writing HTML, the "&" is converted to "&amp;", so in the raw output you'd see "&amp;mdash;". There's no difference in interpretation for text inside or outside inline literals or literal blocks -- there's no character entity interpretation in either case.

It continues talking about a workaround using `|substitution|`, but rST doesn't support nested inline markup which would be needed for that to work in this case (since it's under the "literal" markup of \`\` already)

### [Is nested inline markup possible?](https://docutils.sourceforge.io/FAQ.html#is-nested-inline-markup-possible)

> Not currently, no. It's on the [to-do list](https://docutils.sourceforge.io/docs/dev/todo.html#nested-inline-markup) ([details here](https://docutils.sourceforge.io/docs/dev/rst/alternatives.html#nested-inline-markup)), and hopefully will be part of the reStructuredText parser soon.
> [...]
> There are workarounds, but they are either convoluted or ugly or both. They are not recommended.

I was doing further research and found we could run `html.unescape()` on each file contents from the python standard library, which would enable terms like `&nbsp;`. Probably best to do that as a sphinx extension

JJRcop:
> I was doing further research and found we could run html.unescape() on each file contents from the python standard library, which would enable terms like &nbsp;. Probably best to do that as a sphinx extension

Done

PR #1097, opened 2023-11-02
[doc] build: Add tip for 32-bit installer

JJRcop:
This may help people who really need a 32-bit installer so they can make one for themselves.

There is an argument to be made that those who really need this can find it by reading the `installer.nsi` file and the `makensis` man page, and that may act as a desirable filter; but I figure we are giving them build instructions that they are already following against our warnings, so might as well give them a hand.

Screenshot_20231102_171440

### DIFF
--- a/doc/build.rst
+++ b/doc/build.rst
@@ -300,6 +300,11 @@ Building the Windows installer
 This will build ``looking-glass-host-setup.exe`` under
 ``host/build/platform/Windows/looking-glass-host-setup.exe``
 
+.. tip::
+
+   If you want to build a 32-bit installer, add the ``-DBUILD_32BIT`` flag to
+   ``makensis``
+
 .. seealso::
 
    :ref:`Installing the Host <host_install>`


gnif:
I don't agree with the rationale here:

1) They should not be building a 32 bit version of LG, as such there is no reason to make a 32-bit installer. This makes it seem like it's acceptable to do so.
2) If people are having trouble using this, they should not be using it. Third party builds will be missing the pdb and ivshmem driver. Debugging issues will be more difficult also as people may start distributing the version they "built".

JJRcop:
> 1. They should not be building a 32 bit version of LG, as such there is no reason to make a 32-bit installer. This makes it seem like it's acceptable to do so.

Good point, this is definitely reason enough to not merge.

> 2. If people are having trouble using this, they should not be using it. Third party builds will be missing the pdb and ivshmem driver. Debugging issues will be more difficult also as people may start distributing the version they "built".

I see the concern about people distributing what they built. The instructions we have there already are definitely enough and people who would follow them are hopefully experienced enough to deal with build options like this that we don't intend for people to use. I agree with your points and will close this PR.

PR #1101, opened 2023-11-30
[client] x11: Fix clipboard selection notify

JJRcop:
Was expecting a pointer but getting a pointer-to-pointer

This was causing clipboard sync from host to guest to break on KDE X11


gnif:
Thanks, sorry I missed this PR or I would have merged it :)

PR #1105, opened 2024-01-01
Fixed minor spelling error in logger ("complient" should be spelled as "compliant").

swstegall:
I noticed when running Looking Glass Client that the logger spells the word "compliant" incorrectly. I amended the spelling from "complient" to "compliant".


gnif:
Thanks for this. I will fix this directly as our workflow requires code contributors to also be added to the `AUTHORS` file, and as this is a simple typo, it does not warrant the addition.

PR #1117, opened 2024-04-08
[doc] requirements: Add section encouraging DMABUF use for iGPUs

JJRcop:
Also added link in install_libvirt to the new section.

### DIFF
--- a/doc/requirements.rst
+++ b/doc/requirements.rst
@@ -82,17 +82,16 @@ Linux kernel unless you use the open source NVIDIA drivers.
 iGPUs should use DMABUF
 ^^^^^^^^^^^^^^^^^^^^^^^
 
-While `DMABUF` with the `KVMFR module <ivshmem_kvmfr>` offers performance
+While `DMABUF` with the :doc:`KVMFR module <ivshmem_kvmfr>` offers performance
 benefits for all users, for the often bandwidth-starved users with an iGPU on
-their host it's considered necessary for a decent experience.
+their host it's considered a requirement for a decent experience.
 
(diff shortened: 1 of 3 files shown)


gnif:
The wording needs to change a bit here. There is no "copy engine" per say under Linux, it's a DMA transfer, as such it offloads the transfer from the CPU, to the GPUs DMA hardware.

> without using an intermediary buffer

It might still use one, it depends on the GPU hardware and driver implementation. iGPUs may even avoid any copy at all as they share system RAM with the GPU and could just use it directly without any copy at all.

> considered necessary for a decent experience.

I would reword this to state it is a requirement.

gnif (changes requested):
.

doc/requirements.rst
```diff
@@ -78,3 +76,22 @@ NVIDIA unlike AMD do not seem to suffer from the same stability issues as AMD
...
+
+While `DMABUF` with the :doc:`KVMFR module <ivshmem_kvmfr>` offers performance
+benefits for all users, for the often bandwidth-starved users with an iGPU on
+their host it's considered a requirement for a decent experience.
+
+When using a normal SHM file, many GPU drivers will copy incoming frames from
+shared memory to an intermediary buffer, then upload it from that buffer to the
+GPU's framebuffer. The KVMFR module will instead use Direct Memory Access (DMA)
+to download incoming frames directly from shared memory, which may depending on
+GPU design eliminate the intermediary buffer. This is especially helpful to iGPU
```
JJRcop, line 93: > which may depending on
GPU design eliminate the intermediary buffer.

I hope this modification is correct, this is how I understand it.

PR #1121, opened 2024-05-10
[client] fix some renderer bugs which cause SPICE display corruption

jacobmcnamee:
See individual commits.

Side note: Much of the EGL texture code seems to uniquely use a convention of "stride in pixels, pitch in bytes" despite a comment in `texture.h` stating the opposite. The primary source of corruption resolved by this PR (i.e. incorrect pointer math in `egl_texBufferStreamUpdate`) is only incorrect because the other EGL texture code follows the opposite convention.


client/renderers/EGL/texture_buffer.c
```diff
@@ -196,7 +196,7 @@ static bool egl_texBufferStreamUpdate(EGL_Texture * texture,
     const uint8_t * src = update->buffer;
     for(int y = 0; y < update->height; ++y)
     {
-      memcpy(dst, src, update->pitch);
+      memcpy(dst, src, update->width * texture->format.bpp);
```
gnif, line 199: This change is suspect, please note that we have to handle some rather special formats that are non-standard, such as 24-bit RGB packed into a 32-bit RGBA texture. I'd need to review closer, but I suspect this will cause some issues with these formats.

gnif (commented):
Nice catches, thanks. Yes over the life of this project pitch and stride have been confused, mostly due to my own dyslexia :).

Only one commit here though of concern.

client/renderers/EGL/texture_buffer.c
jacobmcnamee: It was obvious to me that the existing `size = update->pitch` would be incorrect in some cases (e.g. if `update->width` was clamped in caller `egl_textureUpdateRect`).

The `dst` pointer is offset by `update->x * texture->format.bpp` on a previous line, so I assumed this would work.

gnif: You are likely correct here, I just need to do some testing to confirm we don't introduce a bug, there are some nasty and non-obvious edge cases around these routines with regards to these strange formats, and padding for DMA buffer alignment in the guest, and on the host.

gnif: Seems good to me, thanks for the effort on this. Merging

Issue #1129, opened 2024-07-18
Looking glass window crashes randomly on B7-rc1 and B7-rc1-30-d060e375


saqib-saleem:
Looking glass windows is crashing randomly. These are the last three crash reports:
```
05:30:23.618 [E]             crash.c:215  | crit_err_hdlr                  | ==== FATAL CRASH (B7-rc1) ====
05:30:23.618 [E]             crash.c:216  | crit_err_hdlr                  | signal 11 (Segmentation fault), address is 0x7f11f7600000
05:30:23.618 [E]             crash.c:207  | printBacktrace                 | [trace]: (0) /lib/x86_64-linux-gnu/libc.so.6(+0x3d580) [0x7f1245858580]
05:30:23.618 [E]             crash.c:207  | printBacktrace                 | [trace]: (1) looking-glass/looking-glass-B7-rc1/client/build/looking-glass-client(+0x11bb97) [0x5608d7c2db97]
05:30:23.618 [E]             crash.c:207  | printBacktrace                 | [trace]: (2) looking-glass/looking-glass-B7-rc1/client/build/looking-glass-client(+0x11b364) [0x5608d7c2d364]
05:30:23.618 [E]             crash.c:207  | printBacktrace                 | [trace]: (3) looking-glass/looking-glass-B7-rc1/client/build/looking-glass-client(+0x11b6fb) [0x5608d7c2d6fb]
05:30:23.719 [E]             crash.c:202  | printBacktrace                 | [trace]: (4) /home/test/looking-glass/looking-glass-B7-rc1/client/renderers/EGL/texture_framebuffer.c:138 (egl_texFBUpdate)
05:30:23.719 [E]             crash.c:202  | printBacktrace                 | [trace]: (5) /home/test/looking-glass/looking-glass-B7-rc1/client/renderers/EGL/texture.c:179 (egl_textureUpdateFromFrame)
05:30:23.719 [E]             crash.c:202  | printBacktrace                 | [trace]: (6) /home/test/looking-glass/looking-glass-B7-rc1/client/renderers/EGL/desktop.c:418 (egl_desktopUpdate)
05:30:23.719 [E]             crash.c:202  | printBacktrace                 | [trace]: (7) /home/test/looking-glass/looking-glass-B7-rc1/client/renderers/EGL/egl.c:627 (egl_onFrame)
[...]

PR #1136, opened 2024-08-29
[common] rect: fix avx unaligned bytes copy

D0ot:
One extra byte is copied.  it causes SIGSEGV.


D0ot:
fix issue #1129

gnif:
Thanks! well spotted.

As this is such a minor change and you're not in the `AUTHORS` file already, I will make this change directly and credit you for the work.

PR #1149, opened 2024-12-30
[module] MODULE_IMPORT_NS now requires a string literal in 6.13

ticpu:
Reference https://github.com/torvalds/linux/commit/cdd30ebb1

Using a string literal still compiles on 6.11


Svenum:
Some news here? I am waiting for this PR

ticpu:
@netboy3 for what it's worth, I updated my PR, thanks for the helpful review on zeule's PR.

HikariKnight:
@ticpu i would suggest that you use if statements for the kernel module fix for 6.13 as this current patch breaks when building against kernel 6.12.X and older kernels on fedora

```patch
--- a/module/kvmfr.c
+++ b/module/kvmfr.c
@@ -654,7 +654,9 @@ MODULE_DEVICE_TABLE(pci, kvmfr_pci_ids);
 MODULE_LICENSE("GPL v2");
 MODULE_AUTHOR("Geoffrey McRae <geoff@hostfission.com>");
 MODULE_AUTHOR("Guanzhong Chen <quantum2048@gmail.com>");
-MODULE_VERSION("0.0.9");
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,16,0)
+MODULE_VERSION("0.0.12");
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(6,13,0)
+MODULE_IMPORT_NS("DMA_BUF");
+#elif LINUX_VERSION_CODE >= KERNEL_VERSION(5,16,0)
 MODULE_IMPORT_NS(DMA_BUF);
 #endif
```

We at universal blue build the kernel module against multiple kernels for use in our atomic images (with several messages telling people go come to us first with issues since we have to package the module with the image) so users who want to use looking-glass can do so without having to make their own custom atomic image.

And we noticed when we were starting builds against 6.13 and 6.12 with this patch applied, the builds against kernel 6.12 and older would fail, the patch above will make building against older kernels still work while also fixing building against 6.13

ticpu:
> @ticpu i would suggest that you use if statements for the kernel module fix for 6.13 as this current patch breaks when building against kernel 6.12.X and older kernels on fedora

I wonder if it's something specific to Fedora or Arch as this is supposed to be compatible for a long time. Like I said in my first post, it did compile or Arch's 6.11.
Thanks for the tip I'll check it out tonight.

HikariKnight:
> > @ticpu i would suggest that you use if statements for the kernel module fix for 6.13 as this current patch breaks when building against kernel 6.12.X and older kernels on fedora
> 
> I wonder if it's something specific to Fedora or Arch as this is supposed to be compatible for a long time. Like I said in my first post, it did compile or Arch's 6.11. Thanks for the tip I'll check it out tonight.

you are welcome, it is probably a good idea to keep the old behavior on anything that is not 6.13+

ticpu:
Since I was mislead with my v6.11 build on ArchLinux, I decided to test it out with multiple kernels.

Without the patch, anything below 6.13-rc1 fails.

With #if'ed patch, I could build the module on vanilla 6.10, 6.14-rc1, Fedora's 6.12 and ArchLinux's 6.13.

I added both zeule and you as co-author since you both suggested to add the `#if` in the first place.

PR #1158, opened 2025-02-03
[module] fix build on linux 6.13

VMFortress:
Fixing the kernel module namespace import for kernel >=6.13.0 as they now require string literals.


VMFortress:
Did not notice there was already an open PR. Closing.

PR #1161, opened 2025-02-11
[client] overlay/status: don't return damage rect if nothing was drawn

spencercw:
The renderer contains an optimization to skip ImGui if there is nothing to show, but the status overlay always returns a damage rect, even if it didn't draw anything, so the optimization can never take effect.


gnif:
I had to revert this in `b43f572af0c7186597360a74ea82a99972d8ca78` as dragging the overlay window around results in screen corruption to the guest desktop.

PR #1162, opened 2025-02-11
[client] overlay/msg: fix race condition in render

spencercw:
If an overlay is closed with `overlayMsg_close`, the message can be freed while it is still being used by `msg_render`, resulting in a segfault. Lock the message list for the duration of `msg_render` to fix this.

A semi-reliable way to reproduce this is to put a 20ms sleep after the call to `ll_peek_head` in `msg_render`. Use remote desktop to stop the Looking Glass host service, then start the client. The 'host not running' message should appear. Now start the host service. The client will now probably crash somewhere inside `msg_render` because the message it is rendering has been freed by `overlayMsg_close`.

Not sure this is the cleanest approach, but I've taken after the precedence of `ll_forEachNL` by adding some more `nl` methods. Perhaps it would be better to add a new lock in `msg.c`.

### DIFF
--- a/client/src/overlay/msg.c
+++ b/client/src/overlay/msg.c
@@ -85,9 +85,14 @@ static bool msg_needsOverlay(void * udata)
 static int msg_render(void * udata, bool interactive, struct Rect * windowRects,
     int maxRects)
 {
+  ll_lock(l_msg.messages);
+
   struct Msg * msg;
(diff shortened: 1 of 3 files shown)


gnif:
Honestly I am not surprised there is an issue here, the code that handles the messages has not been looked at in quite some time. Thanks for this!

PR #1141, opened 2024-11-04
Add a parameter to hide the waiting message in the GUI

RodoMa92:
Just a small change to hide the popup messagebox in the UI while waiting.

PR #1164, opened 2025-02-16
Add an option to not show the warning popup if the host application is not detected

soyantouwu:
I added this option as a small quality-of-life improvement (since this warning steals focus away from the client and often appears at the worst time), especially when troubleshooting any issue with a non-booting VM or broken GPU, since you can already tell when the host app is not running because of the Spice warning on the top right and the client log.


soyantouwu:
Didn't notice this was already implemented in #1141, closing

PR #1165, opened 2025-02-16
Request changes when authors action fails

JJRcop:
Adds a few steps to the authors job in the pr-check workflow to add a review requesting changes to PRs failing the check.

This review can then be dismissed when they have added themselves.

kuva


JJRcop:
I was going to have a nice demo here but the security of this repo doesn't allow that, which is actually a good thing!
I'm not sure how exactly to fix this, but it is for sure fixable somehow.
Anyway, I will leave the demo commit in for now.

PR #1168, opened 2025-02-21
Fix kvmfr not building with kernel 6.13

AlexanderSchuetz97:
In Kernel 6.13 the MODULE_IMPORT_NS parameter was changed to a string literal. 
I have added a pre processor statement for this case while preserving compatibility with previous kernels.

Sidenote: Not sure why MODULE_VERSION still says 0.0.9? The DKMS config says 0.0.11 Its not been an issue so I did not touch it.


gnif:
Duplicate of #1149, closed.

PR #1174, opened 2025-03-14
[client] pipewire: Stop loading properties from client-rt.conf

netboy3:
Pipewire now automatically moves non-rt clients into non-rt threads so client-rt.conf is obsolete.
Pipewire will now emit a deprecation warning if a client tries to load RT config properties which LG does.
Stop loading RT properties during context initialization for Pipewire 1.3.81 and newer.
See: https://gitlab.freedesktop.org/pipewire/pipewire/-/commit/24bcacc6195ffbf8e40c9ea1374eb6666252eadc

### DIFF
--- a/client/audiodevs/PipeWire/pipewire.c
+++ b/client/audiodevs/PipeWire/pipewire.c
@@ -168,6 +168,12 @@ static bool pipewire_init(void)
   pw_init(NULL, NULL);
 
   pw.loop = pw_loop_new(NULL);
+#if PW_CHECK_VERSION(1, 3, 81)
+  pw.context = pw_context_new(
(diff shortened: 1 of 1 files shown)


spencercw:
More importantly, PipeWire now applies realtime priority automatically under the default configuration, provided the user has appropriate permissions. This may be worth mentioning in the documentation somewhere, because users typically won't have this permission by default. This can be checked with `ulimit -r`: if it's `0` then no realtime. On Arch, realtime can be enabled by installing the `realtime-privileges` package and adding yourself to the `realtime` group.[^1]

[^1]: https://wiki.archlinux.org/title/Realtime_process_management

PR #1200, opened 2025-08-19
[doc] kvmfr: Update udev rule example

netboy3:
Starting with version 258 of systemd, OWNER=/GROUP= are ignored for non-system user/group in udev rules files. Modify the example in the documentation to reflect this change.

### DIFF
--- a/doc/ivshmem_kvmfr.rst
+++ b/doc/ivshmem_kvmfr.rst
@@ -120,12 +120,17 @@ change its ownership manually, i.e.:
 
    sudo chown user:kvm /dev/kvmfr0
 
-As an example, you can create a new file in ``/etc/udev/rules.d/99-kvmfr.rules``
+(replace ``user`` with your username)
+
+As an example, you can create a new file in ``/etc/udev/rules.d/70-kvmfr.rules``
 with the following contents::
 
-   SUBSYSTEM=="kvmfr", OWNER="user", GROUP="kvm", MODE="0660"
+   SUBSYSTEM=="kvmfr", GROUP="kvm", MODE="0660", TAG+="uaccess"
 
-(replace ``user`` with your username)
+.. note::
+
+   Make sure the udev rule file name ordinal value is below (lexically sorts before) ``73-seat-late.rules``
+   to allow the uaccess tag to be processed properly.
 
 .. _ivshmem_kvmfr_libvirt:
 
--- a/doc/words.txt
+++ b/doc/words.txt
@@ -31,6 +31,7 @@ imgui
 ini
 kvmfr
 laggy
+lexically
 libdecor
 libpipewire
 libpulse
@@ -73,6 +74,7 @@ Threadripper
 toolchain
 tritanope
 tunable
+uaccess
 udev
 UEFI
 uncheck


JJRcop:
Version 257 is what is available in Debian 13, which is stable and thus the target of the docs. Does this work in 257?

doc/ivshmem_kvmfr.rst
```diff
@@ -120,12 +120,17 @@ change its ownership manually, i.e.:
...
+As an example, you can create a new file in ``/etc/udev/rules.d/70-kvmfr.rules``
 with the following contents::
 
-   SUBSYSTEM=="kvmfr", OWNER="user", GROUP="kvm", MODE="0660"
+   SUBSYSTEM=="kvmfr", GROUP="kvm", MODE="0660", TAG+="uaccess"
 
-(replace ``user`` with your username)
+.. note::
+
+   Make sure the udev rule file name ordinal value is below ``73-*.rules``
```
JJRcop, line 132: For more accessibility to those who for example speak English as a second language, could this be re-worded a little? Possibly like this (but feel free to adjust it in a way you prefer)


```suggestion
   Make sure the udev rule file name's numerical value is less than 73,
   or in other words up to and including ``72-kvmfr.rules``,
```

doc/ivshmem_kvmfr.rst
```diff
@@ -120,12 +120,17 @@ change its ownership manually, i.e.:
...
    sudo chown user:kvm /dev/kvmfr0
 
-As an example, you can create a new file in ``/etc/udev/rules.d/99-kvmfr.rules``
+(replace ``user`` with your username)
+
+As an example, you can create a new file in ``/etc/udev/rules.d/70-kvmfr.rules``
 with the following contents::
 
-   SUBSYSTEM=="kvmfr", OWNER="user", GROUP="kvm", MODE="0660"
+   SUBSYSTEM=="kvmfr", GROUP="kvm", MODE="0660", TAG+="uaccess"
```
JJRcop, line 128: This works fine on 257, I am using it right now. (As should be expected, since this has been in systemd for 13 years)

JJRcop (approved):
This is a good change and could be merged as-is, but I have a non-blocking request for changes regarding the note.

doc/ivshmem_kvmfr.rst
JJRcop: The reason I have suggested to insist on ``xx-kvmfr.rules`` instead of ``xx-*.rules`` is even though the wildcard is more technically correct and may properly inform users that they can use whatever name they want, advanced users who already know how these rules work will understand they can do that anyway, where as novices may become confused.

Even if they misunderstand that it must be named ``xx-kvmfr.rules``, that's ok.

netboy3: > For more accessibility to those who for example speak English as a second language, could this be re-worded a little?

Your suggestion is technically inaccurate. I'm not comfortable in modifying text to be inaccurate just for the sake of catering to ESL users. The rule file name does not have to have a "numerical value" as you suggested. In fact, the only requirement for a rule file name is to have a `.rules` extension. The file `+kvmfr.rules`, for example, will work just fine and will properly precede the `uaccess` processing rule.

Would you be more comfortable in using "lexical order precedes" instead of "ordinal value is below"?

> The reason I have suggested to insist on xx-kvmfr.rules instead of xx-*.rules is even though the wildcard is more technically correct and may properly inform users that they can use whatever name they want, advanced users who already know how these rules work will understand they can do that anyway, where as novices may become confused. Even if they misunderstand that it must be named xx-kvmfr.rules, that's ok.

Again, I'm not comfortable stating inaccuracies, especially in technical documentation. I can similarly reason that such statements will confuse knowledgeable users and might even trigger them to question our ability to understand the underlying system. From my experience in assisting our users, most of them don't even read the "Notes" boxes. They will copy and paste the suggested file name (`70-kvmfr.rules`) above the box and move on.

JJRcop: On reflection, use of parenthesis for clarification, ensuring the precise technical explanation remains, might be enough. For example
```suggestion
   Make sure the udev rule file name ordinal value is below (sorts before) ``73-*.rules``
```

JJRcop (approved):
Looks great. My concerns were elegantly resolved in 0431c70cf4c0e46609e5e76bb3f16248b59a9f88 👍

steawth:
Requesting these changes be pushed into the live docs site. My distro has updated their systemd package past where this change occurs so others will likely run into the same issue with the udev rules when following the current docs installation guide.

netboy3:
@steawth, this PR was merged 4 months ago. Once a documentation change is committed, it appears in the "live docs site" within a few seconds. The documentation is an integrated part of the project and is version controlled similarly to the rest of it. Every project commit is generating a new set of documentation that is available on the documentation site. You can easily get to links of notable versions of the documentation on the [project download page](https://looking-glass.io/downloads). If you want to read the latest version of the documentation, you can reach it at the [Bleeding Edge](https://looking-glass.io/docs/bleeding) version. Obviously, if you're running stable, don't expect bleeding-edge features to work (which is why the documentation version is in sync with the project itself).

steawth:
ah I see, I didn't know there was a bleeding edge version of the docs site. sorry for the ping and thanks for linking it

JJRcop:
@steawth @netboy3 Documentation targets Debian Stable, which currently uses systemd v257, just before this change is needed.

If Debian Stable changes their configuration between stable releases of Looking Glass that would require changes to the documentation, we may consider releasing a new version with updated documentation (not based off of the master branch, but off of the last released stable version), for example maybe B7.1.

However, as the changes on the master branch regarding udev rules does not currently affect Debian Stable, the target system, no interim updates are needed at this time.

PR #1207, opened 2025-09-12
[idd] all: delete ARM and ARM64 targets

quantum5:


RadxaYuntian:
I'm preparing a virtualized Windows on Arm demo for our company's product (we mainly make Arm based SBCs), and though I might want to check what LG is doing these days, as I remember seeing gnif posted something in discord about frame streaming without GPU. Only to find out this is removed (assuming this indirect display driver is the enabler for that usage).

Let us know if there is something we can help to bring this to Arm as well.

PR #1230, opened 2025-09-14
 [idd] helper: add icon context menu with log directory open

quantum5:
Also use a better algorithm to get `ProgramData` instead of hard-coding.


gnif:
Manually merged

PR #1146, opened 2024-12-17
Wayland: Mouse only scrolls up

neonbyte1:
I encountered a problem with the mouse scrolling when switching from X11 + i3 to Wayland + Hyprland. At first I thought that this might be due to an outdated version of Looking Glass, as I was still using version B6 in my gaming VM. However, I performed a clean installation of my system last night, set up a new VM , but the problem was still there - even with the bleeding edge version.

## Actual problem

I've noticed that the actual value of the `value` parameter in `pointerAxisHandler` is sometimes `0` while scrolling up **or** down. After looking at the implementation, you'll see why.

```C
int button = value > 0 ?
  5 /* SPICE_MOUSE_BUTTON_DOWN */ :
  4 /* SPICE_MOUSE_BUTTON_UP */;
app_handleButtonPress(button);
app_handleButtonRelease(button);
```

Since the `value` is not greater than `0`, the LookinGlass client will always handle the event as `SPICE_MOUSE_BUTTON_UP`.

### Preview

https://i.imgur.com/A7eGpMb.mp4

### Logs

```
Serial: 7067714, Axis: 0, Value: 0, Double: 0.000000
Serial: 7067716, Axis: 0, Value: 0, Double: 0.000000
Serial: 7067717, Axis: 0, Value: 0, Double: 0.000000
Serial: 7067719, Axis: 0, Value: 3840, Double: 15.000000
Serial: 7067720, Axis: 0, Value: 0, Double: 0.000000
Serial: 7067722, Axis: 0, Value: 0, Double: 0.000000
Serial: 7067724, Axis: 0, Value: 0, Double: 0.000000
Serial: 7067729, Axis: 0, Value: 0, Double: 0.000000
Serial: 7067732, Axis: 0, Value: 0, Double: 0.000000
Serial: 7067737, Axis: 0, Value: 0, Double: 0.000000
Serial: 7067740, Axis: 0, Value: 0, Double: 0.000000
Serial: 7067743, Axis: 0, Value: 3840, Double: 15.000000
Serial: 7067808, Axis: 0, Value: -3840, Double: -15.000000
Serial: 7067810, Axis: 0, Value: 0, Double: 0.000000
Serial: 7067812, Axis: 0, Value: 0, Double: 0.000000
Serial: 7067814, Axis: 0, Value: 0, Double: 0.000000
Serial: 7067818, Axis: 0, Value: 0, Double: 0.000000
Serial: 7067829, Axis: 0, Value: -3840, Double: -15.000000
[...]

PR #1256, opened 2026-01-07
Fix for downward scroll wheel misbehaviour when using the Wayland client

afiretruck:
When using the Wayland client, scrolling down with the mouse wheel resulted in 1x scroll down command being sent to the VM, followed by 3x scroll up commands. Scrolling up on the mouse wheel worked as expected.

I tracked this down to the ``pointerAxisHandler`` function inadvertently sending ``SPICE_MOUSE_BUTTON_UP`` events when the input ``value`` was 0. 

This fix wraps the ``int button = value > 0 ?`` test with a ``value != 0`` test.

### DIFF
--- a/client/displayservers/Wayland/input.c
+++ b/client/displayservers/Wayland/input.c
@@ -91,11 +91,15 @@ static void pointerAxisHandler(void * data, struct wl_pointer * pointer,
   if (axis != WL_POINTER_AXIS_VERTICAL_SCROLL)
     return;
 
-  int button = value > 0 ?
-    5 /* SPICE_MOUSE_BUTTON_DOWN */ :
-    4 /* SPICE_MOUSE_BUTTON_UP */;
-  app_handleButtonPress(button);
-  app_handleButtonRelease(button);
+  if (value != 0)
+  {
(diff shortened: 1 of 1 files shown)


afiretruck:
I should add - I'm testing this with Hyprland.

gnif:
Closing, duplicate of #1146
We will not be adding a workaround for Hyprland doing stupid things, events with a delta of `0` are a no-op and should not be sent to applications.

afiretruck:
While I totally understand your point of view - Hyprland _is_ doing something silly here - I doubt Hyprland is the only one doing so now, and in the future. You can't guarantee the environment looking glass will run in is going to be 100% perfect, Hyprland or not. Everything has its quirks.

Perhaps you could reconsider? What harm is there in a little bit of defensive programming?

gnif:
Hyprland absolutely is the only one doing so. If we added a quirk to our code for every time Hyprland broke something our code would be a mess. Hyprland is notorious for not following the standards and having issues with Looking Glass as a result, which is why we do not recommend the usage of Hyprland as a DE in general.

PR #1257, opened 2026-01-07
Use double type for storing the display scaling factor in Wayland

afiretruck:
One of my monitors has a resolution of 3840 x 2560. I have this set up with a scaling factor of 1.6x to achieve a comfortable widget size akin to that of a 30" 4k display at 1.5x scale. Unfortunately the ``wl_fixed`` type stores 1.6 as 1.6034somethingsomething (I can't remember the exact number). This results in an inaccurate scaled resolution (an additional 3 or so pixels in each dimension) on my funky monitor. 

The fix is to replace the ``wl_fixed`` typed with ``double`` in the wayland display server.

I'm not sure if there's a better solution to this problem. Please advise if there is one - I'd be happy to implement it if so.

### DIFF
--- a/client/displayservers/Wayland/cursor.c
+++ b/client/displayservers/Wayland/cursor.c
@@ -183,7 +183,7 @@ void waylandCursorFree(void)
 
 void waylandCursorScaleChange(void)
 {
-  int newScale = ceil(wl_fixed_to_double(wlWm.scale));
+  int newScale = ceil(wlWm.scale);
   if (newScale == wlWm.cursorScale)
     return;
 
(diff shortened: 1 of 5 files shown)


gnif:
While the fix is sound, the implementation is not. The main concern I had here was the equality check on L49. Floating point values are not exact values, as such equality comparisons are never reliable. Instead I have implemented this using a rational scale representation, storing the scale as a numerator/denominator pair.

This keeps the fractional scale exact for cases like 3840 / 2400, avoids the precision loss from wl_fixed_t, and also avoids introducing floating-point comparisons into the Wayland state handling. The value is only converted to double at the point where the app resize API requires it.

PR #1259, opened 2026-01-16
Fix discrete mouse scroll speed on Wayland

Janrupf:
### Mission Impossible: Mouse scrolling on Wayland

The LookingGlass client interpreted every `wl_pointer.axis` event as "mouse wheel has turned one step". On a traditional mouse this is true, but with mice that have high resolution scroll wheels this event is sent multiple times:

"Traditional" mouse:
```
// move wheel
// click
Scroll delta: 15.0
// move wheel
// click
Scroll delta: 15.0
// move wheel
// click
Scroll delta: 15.0
```

High resolution mouse:
```
// move wheel
Scroll delta: 3.0
Scroll delta: 7.0
Scroll delta: 2.0
// click
Scroll delta: 3.0
// move wheel
Scroll delta: 5.0
Scroll delta: 4.0
Scroll delta: 1.0
// click
Scroll delta 5.0
```

The way that Looking Glass originally interpreted this lead to a funny behavior: The faster you moved the scroll wheel on a high resolution mouse, the less the scroll distance the client recognized because the Compositor consolidated multiple events into one. I'm not 100% sure that a step size of `15.0` is the correct value, but the wayland spec reads

> For scroll events (vertical and horizontal scroll axes), the value parameter is the length of a vector along the specified axis in a coordinate space identical to those of motion events, representing a relative movement along the specified axis.

For mice that is absolutely ambiguous, because no one knows how the amount of turning a scroll wheel relates to moving the cursor, but this seems to commonly be interpreted as 15px = one line (which as far as I can tell is the constant from X11).

Looking Glass could make use of `wl_pointer.value120` events, but they have their own ambiguity in the Wayland spec and there is no real advantage, because SPICE doesn't support discrete scroll events at this point in time.

### How this relates to Hyprland scrolling

#1256 and https://github.com/gnif/LookingGlass/pull/1146 have attempted to fix weird scroll behavior on Hyprland by ignoring zero values. The reason that Hyprland sends zero values for axis scroll values is because it attempts to work around bugs in applications not properly handling `wl_pointer.value`, hoping that values of zero are ignored. It has to send the `wl_pointer.value` event for applications that handle `wl_pointer.value120`, because the spec requires it do so so... maybe, depending on who you ask. 

So the workaround in Hyprland causes scrolling up - disabling the workaround in Hyprland with `input:emulate_discrete_scroll = 0` fixes the issue then, right? Well, yes, kind of. LookingGlass doesn't scroll the wrong way anymore, and Hyprland doesn't send zero values, instead it actually sends whatever the hardware generates. Then LookingGlass wrongly assumes that a single `wl_pointer.axis` event means the mouse wheel turned one full step, causing way too fast scrolling.

This is a combination of a bug in LookingGlass' input handling and Hyprland's workaround not quite working (it does work for Java actually, there Hyprland's workaround fixes scrolling in Java) - in the end caused by the Wayland spec being unclear.


gnif:
Thanks for this. I would be happy to merge this but the fix for the uninitialized memory should not be part of this PR.

I am not sure how this can even occur, `msg` is never used if `lgmpClientProcess` doesn't return LGMP_OK, and `lgmpClientProcess` will never return `LGMP_OK` without initializing it.
https://github.com/gnif/LGMP/blob/cee29332212b40f54a57cf36e39d301c02c0ce07/lgmp/src/client.c#L442-L446

Also be sure to add yourself to the AUTHORS file.

Janrupf:
I'm not sure about the memory corruption, I've removed that patch for now. I just noticed it because my LookingGlass client was crashing when run from the IDE and adding that fixed it, but I honestly have no idea why.

PR #1275, opened 2026-06-03
[idd] driver: add BOM to CEdid.* for non-English systems

quantum5:
It wouldn't compile otherwise.


gnifAMD:
huh, strange, even the CI server was able to build it. Perhaps a different version of VS?

quantum5:
> huh, strange, even the CI server was able to build it. Perhaps a different version of VS?

It's something to do with code pages. If the default system code pages contains ©, then it will compile without BOM. Otherwise, it won't.

PR #1277, opened 2026-06-03
[idd] helper: implement warning when there is no GPU

quantum5:
This PR implements the following balloon notification when no GPU is detected:

image

This is defined as not having any display device from AMD, Intel, or Nvidia.

The warning can be disabled by toggling the checkbox:

image

### DIFF
--- a/idd/LGIddHelper/CCheckbox.cpp
+++ b/idd/LGIddHelper/CCheckbox.cpp
@@ -1,0 +1,38 @@
+﻿/**
+ * Looking Glass
+ * Copyright © 2017-2026 The Looking Glass Authors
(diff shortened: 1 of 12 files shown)


gnif:
I think we should have the driver itself report if it wasn't backed by a GPU, which can occur when the device is there but it's in a broken state

quantum5:
> I think we should have the driver itself report if it wasn't backed by a GPU, which can occur when the device is there but it's in a broken state

Do you have any specific API in mind?

gnif:
Perhaps just use the named pipe to tell the helper there is no GPU acceleration.

PR #1282, opened 2026-06-03
[client] wayland: use wp_fractional_scale_v1 for pixel-exact output

serrj-sv:
# [client] wayland: use wp_fractional_scale_v1 for pixel-exact output

## Problem

On a fractionally-scaled Wayland session (e.g. GNOME/Mutter at 125 %, 150 %,
175 %) the EGL/OpenGL output is slightly soft and geometrically distorted
instead of producing a crisp 1:1 image. An SDL3-based client such as FreeRDP 3's
SDL frontend (`sdl-freerdp3`) is pixel-perfect on the same setup, which is the
behaviour this PR brings to Looking Glass.

## Root cause

The Wayland EGL backend renders into a `wp_viewport`, but it never asks the
compositor what scale the surface is actually displayed at. Two issues combine:

1. **The scale is guessed from output geometry** (`output.c`):
   ```c
   node->scale = waylandScaleFromRatio(modeWidth, node->logicalWidth);
   ```
   `logical_width` is itself rounded by the compositor (`round(mode / scale)`),
   so reversing the division cannot recover the exact `n/120` fraction the
   compositor uses for the surface. The `WaylandScale` rational type added in
   01b8724c makes this division lossless, but the *input* has already lost
   information, so the result is still an approximation.

   Example — 2560 px mode at 150 %:
   ```
   logical_width = round(2560 / 1.5) = 1707     (2560 / 1.5 = 1706.67)
   derived scale = 2560 / 1707       = 1.50029...   (not 1.5)
   ```

2. **The buffer is sized with a truncating multiply** (`gl.c`):
   ```c
   wl_egl_window_resize(wlWm.eglWindow,
       waylandScaleMulInt(wlWm.scale, width),   // floor
       waylandScaleMulInt(wlWm.scale, height), 0, 0);
   ```
   Combined with `wp_viewport_set_destination(width, height)`, a buffer that is
   even a pixel short of `logical * scale` gets stretched to fill the
   destination — a non-integer resample of the whole frame.

The net effect is that the buffer no longer maps 1:1 to physical pixels and the
compositor resamples it.

## Fix

* Bind **`wp_fractional_scale_v1`** and create a per-surface object. Its
  `preferred_scale` event reports the compositor's exact scale as a numerator
  over 120, which maps directly onto the existing `WaylandScale` type:
  ```c
  wlWm.preferredScale = waylandScaleFromRatio(scale, 120);
  ```
  `waylandWindowUpdateScale()` prefers this authoritative value when available.

* Add **`waylandScaleMulIntCeil()`** (integer-only, mirrors the existing
  `waylandScaleMulInt()`) and size the EGL buffer with it, so
  `ceil(logical * scale)` exactly covers the viewport destination.

With the exact scale and a matching buffer size, Mutter takes its no-resample
path and the image is pixel-exact.

## Compatibility / scope

* Gated behind the existing `wayland:fractionScale` option (default on) and only
  active when both `wp_viewporter` and `wp_fractional_scale_manager_v1` are
  advertised.
* Compositors without `wp_fractional_scale_v1` fall back to the existing
  output-geometry heuristic — **no behavioural change** there.
* Integer scales (`preferred_scale == 120, 240, …`) continue to take the
  `wl_surface_set_buffer_scale()` path.

## Note on the protocol file

`fractional-scale-v1.xml` is bundled locally under
`client/displayservers/Wayland/protocol/` because the pinned
`repos/wayland-protocols` submodule (v1.25) predates the protocol, which was
added upstream in wayland-protocols **1.31**. The CMake generator points at the
local copy.

Alternative, if preferred by maintainers: bump the `repos/wayland-protocols`
submodule to ≥ 1.31 and reference `staging/fractional-scale/fractional-scale-v1.xml`
from `${WAYLAND_PROTOCOLS_BASE}` like the other protocols, dropping the local
copy. Happy to switch the PR to that approach.

## Testing

Builds and links cleanly against current `master` (`b3701992`) with no new
warnings. Verified on GNOME/Mutter under fractional scaling: output is sharp
with no geometric distortion; the `wp_viewport` destination (logical size) and
the `ceil(logical * scale)` buffer size produce a 1:1 mapping to physical pixels.

### DIFF
--- a/client/displayservers/Wayland/gl.c
+++ b/client/displayservers/Wayland/gl.c
@@ -92,8 +92,14 @@ void waylandEGLSwapBuffers(EGLDisplay display, EGLSurface surface, const struct 
 
     int width, height;
     wlWm.desktop->getSize(&width, &height);
-    wl_egl_window_resize(wlWm.eglWindow, waylandScaleMulInt(wlWm.scale, width),
-        waylandScaleMulInt(wlWm.scale, height), 0, 0);
+
+    // Size the buffer to ceil(logical * scale) so it exactly covers the
+    // viewport destination. Truncating (floor) would leave the buffer up to a
+    // pixel short and force the compositor to stretch (and thus resample) the
+    // image, producing the distortion seen with fractional scale.
+    wl_egl_window_resize(wlWm.eglWindow,
+        waylandScaleMulIntCeil(wlWm.scale, width),
+        waylandScaleMulIntCeil(wlWm.scale, height), 0, 0);
 
     if (width == 0 || height == 0)
       skipResize = true;
--- a/client/displayservers/Wayland/protocol/CMakeLists.txt
+++ b/client/displayservers/Wayland/protocol/CMakeLists.txt
@@ -39,6 +39,11 @@ wayland_generate(
 wayland_generate(
   "${WAYLAND_PROTOCOLS_BASE}/stable/viewporter/viewporter.xml"
   "${CMAKE_BINARY_DIR}/wayland/wayland-viewporter-client-protocol")
+# fractional-scale-v1 is bundled locally because the pinned wayland-protocols
+# submodule predates it (it was added upstream in wayland-protocols 1.31).
+wayland_generate(
+  "${CMAKE_CURRENT_SOURCE_DIR}/fractional-scale-v1.xml"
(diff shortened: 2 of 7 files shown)


gnif:
Superseded by #1286. Thanks for the heads up on this.

PR #1295, opened 2026-06-10
[host] D12: fix ACCESS_VIOLATION crash in d12_enumerateDevices with NVIDIA driver 32.x

MatteoGalet:
With NVIDIA driver 32.0.15.9636 (and likely other 32.x releases),
IDXGIAdapter1_EnumOutputs() returns an error code that is neither S_OK
nor DXGI_ERROR_NOT_FOUND. The original loop condition
`!= DXGI_ERROR_NOT_FOUND` treated any such error as "keep iterating",
executing the loop body with *output == NULL. This caused an
ACCESS_VIOLATION crash at IDXGIOutput_GetDesc(*output, ...) since
*output is NULL.

Fix:
- Change the loop condition from `!= DXGI_ERROR_NOT_FOUND` to `== S_OK`
  so iteration only continues on full success.
- Add a defensive null guard `if (!*output) break;` before
  IDXGIOutput_GetDesc to prevent the crash even if the condition is
  ever bypassed.

Tested on:
- GPU: NVIDIA T550 Laptop GPU (10de:1fb7)
- Driver: 32.0.15.9636
- OS: Windows 11 24H2
- LG: B7

### DIFF
--- a/host/platform/Windows/capture/D12/d12.c
+++ b/host/platform/Windows/capture/D12/d12.c
@@ -989,9 +989,10 @@ static bool d12_enumerateDevices(
 
     for(
       int n = 0;
-      IDXGIAdapter1_EnumOutputs(*adapter, n, output) != DXGI_ERROR_NOT_FOUND;
+      IDXGIAdapter1_EnumOutputs(*adapter, n, output) == S_OK;
       ++n, comRef_release(output))
     {
+      if (!*output) break;
       IDXGIOutput_GetDesc(*output, &outputDesc);
 
       if (optOutput)


gnif:
1. Please fix commit summary to a maximum of 50 characters.
2. Confirm if this was authored by you or an AI agent as the description looks very AI generated. If it was authored by you, please add yourself to the `AUTHORS` file.
3. Fix the null guard to follow the formatting of other single line if statements.
4. When there is a result that is not `S_OK` and not `DXGI_ERROR_NOT_FOUND` it should be logged as a warning, along with what the error was to aid in future debugging.

quantum5:
I'll also note that comparing with `S_OK` is not correct in COM. The `SUCCEEDED` macro should be used.

gnif:
@quantum5 feel free to make a PR that handles this correctly if you like, otherwise I will when I get time.

gnif:
Fixed in #1296

In future, do not submit AI pull requests, we will not accept them. If you find an issue, report it, along with the technical details and we'd be happy to implement it. AI PRs just make our job harder as they are often overly verbose and/or completely miss the mark when it comes to the fixes.

PR #1296, opened 2026-06-17
[host] d12: handle IDXGIAdapter::EnumOutputs errors

quantum5:
Before, the code treated any errors other than `DXGI_ERROR_NOT_FOUND` as if they were successes, which is incorrect.

I haven't actually tested this code since I don't have a VM with the host application these days, only the IDD.


gnif:
LGTM

PR #1322, opened 2026-07-30
[client] overlay: load Latin, Greek, and Cyrillic into font atlas

quantum5:


gnif:
Dynamic atlas reloading was added instead. Thanks :)

PR #1324, opened 2026-08-01
[client] app: add minimize on capture release option

trigili:
## Summary

- add the opt-in `win:minimizeOnCaptureRelease` client option
- minimize the client when the configured escape key changes capture mode from enabled to disabled
- document the new option in the client usage reference

## Motivation

When Looking Glass is used like a fullscreen guest application, releasing input capture leaves the client focused and requires a separate host-side app switch. This option lets the window manager restore the previous host application when the escape key releases capture.

The option defaults to disabled, so existing behavior is unchanged. It is limited to the explicit escape-key capture toggle and does not alter focus-loss or overlay behavior.

## Validation

- built the current `master` client successfully with CMake (Wayland, X11, EGL, OpenGL, and PipeWire enabled)
- confirmed `win:minimizeOnCaptureRelease` appears in the generated `--help` output
- ran `git diff --check`
- manually validated the equivalent behavior in a B7-based client on GNOME Wayland before porting it to current `master`


client/src/app.c
```diff
@@ -518,7 +518,12 @@ void app_handleKeyReleaseInternal(int sc)
     {
       if (!g_state.escapeHelp && g_params.useSpiceInput &&
           !app_isOverlayMode())
-        core_setGrab(!g_cursor.grab);
+      {
+        const bool enable = !g_cursor.grab;
+        core_setGrab(enable);
```
gnif, line 523: `enable` is redundant. Please do not submit PRs from LLMs that have not been properly reviewed in future.

gnif (changes requested):
Looks good to me except the obvious LLM involvement without review. Once the issue is fixed, be sure to add yourself to the `AUTHORS` file.

client/src/app.c
unatried: I believe `enable` is just there for readability, since it's being used as an if condition later on aswell.

gnif:
Closing due to lack of response from the author

PR #1325, opened 2026-08-03
client/wayland: fix double xdg_toplevel role assignment in libdecor path

cristatus:
libdecor_decorate() already creates an xdg_surface and xdg_toplevel internally for the frame. Calling xdg_surface_get_toplevel() again on that same xdg_surface just to fetch the toplevel for icon handling assigns the xdg_toplevel role to the underlying wl_surface a second time, which is a protocol violation.

On GNOME Shell 50 (mutter) this is no longer tolerated: mutter terminates the Wayland connection ("WL: error in client communication"), and since libdecor_shellInit() waits for the initial configure with a blocking, error-less libdecor_dispatch(-1) loop, the client hangs forever with no window and no diagnostic output. This made -DENABLE_LIBDECOR=ON (the default when libdecor-0 is present) appear completely broken on current GNOME/Wayland.

Use libdecor_frame_get_xdg_toplevel() instead, which returns the xdg_toplevel libdecor already owns, without requesting a second role assignment.

Require libdecor-0 >= 0.1.1, which is available in Debian oldstable, so the new accessor is guaranteed by the build dependency.


gnif:
Thankyou for debugging this.

Can you please alter your commit to follow the project's convention:

`[client] wayland: ...`

Also please add yourself to the AUTHORS file

PR #1328, opened 2026-08-11
[client] Fix client crashes from cross-thread Wayland resizes and overlay teardown

cristatus:
Two shutdown crashes, both from code running on the wrong thread:

- `waylandSetWindowSize` called `shellResize()` directly from the render,
  frame, and transport threads. Shell requests — and with libdecor-gtk,
  pango decoration rendering — are only safe on the event-loop thread;
  exiting threads crashed in pango's fontmap teardown. Resizes are now
  published atomically and applied by the event loop via an eventfd.

- The render thread freed the overlays on exit while the main thread could
  still be dispatching input callbacks that iterate them. They are now
  freed in `lg_shutdown()`, after the render thread is joined and the
  display server has shut down.

### DIFF
--- a/client/displayservers/Wayland/wayland.c
+++ b/client/displayservers/Wayland/wayland.c
@@ -119,6 +119,8 @@ static bool waylandInit(const LG_DSInitParams params)
   atomic_init(&wlWm.cmCanDoHDR, false);
   atomic_init(&wlWm.hdrPQWhiteLevel, 203);
   atomic_init(&wlWm.hdrScRGBWhiteLevel, 80);
+  atomic_init(&wlWm.pendingResize, 0);
+  wlWm.resizeEventFd = -1;
 
   wlWm.display = wl_display_connect(NULL);
   if (!wlWm.display)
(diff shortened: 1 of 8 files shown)


gnif:
The Wayland resize approach looks correct, but there are two blocking teardown issues:

1. `lg_shutdown()` now frees overlays even if `app_initOverlays()` was never called. If startup fails before the render thread initializes them, `splash_free()` calls `stringlist_free()` on its uninitialized `NULL` lists and crashes. Please track overlay initialization or make teardown explicitly safe for partial initialization.

2. Overlay destruction still occurs before all input callbacks have stopped. On X11, `shutdown()` does not join the event thread; that happens in `free()`, which currently runs after the overlays are destroyed. The evdev thread can similarly still be processing an input batch. Both can call `app_isOverlayMode()` and access the freed overlay list. Please join these threads before freeing overlays, move overlay destruction after `g_state.ds->free()` and ensure `evdev_stop()` has completed.

There is also a minor failure-path leak: if Wayland initialization fails after registering the resize eventfd, it is never unregistered or closed. Please add cleanup for the new event source on initialization failure.

The atomic/eventfd resize coalescing itself looks sound and should fix the original cross-thread Wayland resize problem.

gnif (changes requested):
as per the comment

gnif:
Looks good, thanks!

PR #1329, opened 2026-08-12
[client] app: free the frame event after display server teardown

cristatus:
Display server and overlay teardown can still call app_invalidateWindow, which signals the frame event; the Wayland presentation module does so when it unregisters its graph. Freeing the event beforehand asserted in lgSignalEvent while closing the client.


gnif:
Thanks, I saw this too and had not yet debugged.

PR #1330, opened 2026-08-12
[client] wayland: apply the surface scale before the first commit

cristatus:
The buffer scale and viewport were only set from the resize handling in the swap path, which runs after the buffer has been committed. The first commit therefore carried a buffer of scale x the window size with no scale declared, so the compositor sized the window from the raw buffer. On a scaled output that is larger than the work area GNOME maximizes it, after which libdecor drops every resize as the frame is no longer floating, leaving the window stuck at the size of the display.

Split the scaling out of the swap path and apply it when the EGL window is created, so it is in place for the first commit.

### DIFF
--- a/client/displayservers/Wayland/gl.c
+++ b/client/displayservers/Wayland/gl.c
@@ -35,6 +35,40 @@
 #include "egl_dynprocs.h"
 #include "eglutil.h"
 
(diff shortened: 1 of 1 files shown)


gnif:
LGTM, thanks

PR #1337, opened 2026-09-14
[client] wayland: restore floating size after fullscreen

cristatus:
Remember the last floating content size in both desktop backends and restore it when leaving fullscreen. Keep explicit XDG window geometry synchronized with subsequent configure sizes.


gnif:
Looks good but please wrap your commit message at 74 characters.

PR #1338, opened 2026-09-14
[client] input: add a hold key to disable guest input

babbaj:
I find that needing to toggle input with scroll lock+I to alt+tab in and out of looking glass to be very inconvenient.  This adds a new key that when held disables input for the purpose of switching windows. When looking glass is focused again input is automatically enabled. I chose pause because it's right next to scroll lock.

If input was disabled with scroll lock+I pressing pause will have no affect and focusing looking glass will not automatically enable input so the behavior of toggling input has not changed.


babbaj:
not sure why the authors check is failing, im already in there

gnif:
Sorry but this feature is both confusing and seems very niche to your use case. I appreciate the effort but I don't think we will be adding this.

